PR #18, opened 2022-08-08
Add default placeholder for post UI :)

jcarucci27:
Found it a little confusing to locate the "editor" frame while using. Made this hack since there's no placeholder for SwiftUI yet...


jb55:
Ah I was trying to find a way to do this...

There seems to be one issue where if you start typing before you click anywhere it won't remove the message. Other than that LGTM!

I cherry-picked the carousel change so thanks for that!

jcarucci27:
@jb55 check out my PR little sloppy but that's what happens when Apple forces us to "roll our own" placeholder lol

jb55:
cherrypicked into 5020ae7e3c0b154f0a68ac2bb386786a15acd3d5 with a few tweaks! thanks!

Issue #19, opened 2022-08-08
Need cancel button after search


jb55:
We need a little (x) button on the far right as way to reset the search box

jb55:
Fixed in 7126a68723962fe8bbe020e40622b56142274e74

jb55:
Fixed in d6ef1c40197c7bfe716942a9062dcd5864fd1343

PR #21, opened 2022-08-09
Placeholder text for DM channels with 0 messages

jcarucci27:


jcarucci27:
@jb55 also added a fix for #19

jb55:
looks like this needs rebased, I see a bunch of old commmits on here

Issue #23, opened 2022-08-09
A way to filter out replies in the home timeline


jb55:
Sometimes it gets too chatty.

jcarucci27:
@jb55 I'm taking a stab at this. What I can't figure out right now is how to tell if a given event is a reply and thus should be filtered. But I'm damn close. Check out https://github.com/jcarucci27/damus/commit/9a7f009f88355fb8d0da84ebca605f8819cd228e most specifically line 119 where I think a filter on your events may do the trick.

jb55:
On Wed, Aug 10, 2022 at 05:25:35PM -0700, jcarucci27 wrote:
> What I can't figure out right now is how to tell if a given event is a reply and thus should be filtered.

I'm reworking some of this logic now, I will test this soon

jb55:
lol wtf, I sent the previous message a month ago.

PR #24, opened 2022-08-12
Starting the filters view

jcarucci27:
I don't expect this to get approved but I created a UI that lets you toggle a variable that can conceivably help you filter out replies. Check my comments on issue #23

### DIFF
--- a/damus/Views/ChatView.swift
+++ b/damus/Views/ChatView.swift
@@ -14,6 +14,7 @@ struct ChatView: View {
     
     let damus_state: DamusState
     
+    @State var expand_reply: Bool = false
     @EnvironmentObject var thread: ThreadModel
     
     var just_started: Bool {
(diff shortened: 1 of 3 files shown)


jb55:
looks like this needs a rebase

jcarucci27:
Agreed. It isn’t even finished yet was frying to get your advice on how to
discern if an event is a reply. Once I have that I’ll write the code and
rebase

jb55:
On Fri, Aug 12, 2022 at 03:28:36PM -0700, jcarucci27 wrote:
>Agreed. It isn’t even finished yet was frying to get your advice on how to
>discern if an event is a reply. Once I have that I’ll write the code and
>rebase

yeah it's just quicker for me to test the branch and suggest the proper
behavior if I have a working branch to interact with.

jb55:
what the... did github get confused here? it shows as merged

jcarucci27:
My guess is that it does that cuz I rebased to what you were already
shipping which seems silly in this case but there may be good mechanics
reasons why they do that.

I can resubmit a new one when I get home tonight

PR #25, opened 2022-08-13
Add filters for home screen 

jcarucci27:
Rebased. Relates to #23

### DIFF
--- a/damus/ContentView.swift
+++ b/damus/ContentView.swift
@@ -59,6 +59,8 @@ struct ContentView: View {
     @State var profile_open: Bool = false
     @State var thread_open: Bool = false
     @State var search_open: Bool = false
+    @State var filters_showing : Bool = false
+    @State var should_show_replies : Bool = true
     @StateObject var home: HomeModel = HomeModel()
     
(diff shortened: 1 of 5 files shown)


jb55:
Thanks! only took a small tweak to make it work. Cherry-picked into: 12eb50255af20fc33898b03dd820f0de352dc627

jcarucci27:
Was just checking glad it worked out

*chuck Norris thumbs up*

jb55:
Re-opening this for now. after playing with it for a bit I think we would need a few things before this is properly merged:

- Top-tab view like twitter that switches between "Posts" and "Posts & Replies" instead of the filter drawer. Maybe we can take a look at the filter drawer again if we have even more filter conditions...
- StartBlock (opposite of EndBlock) which provides a consistent scroll target. Right now the top event could get filtered out and then scrolling to top stops working

jcarucci27:
Ok the term for the first bullet is called a segmented control in classic
swift I’ll look into how to do that

For the second one yea I think the scroll target has to just be dynamic and
react to the Boolean we already have nothing major there

PR #30, opened 2022-09-04
Take another stab at #23

jcarucci27:
Took a stab at #23 using the segmented control paradigm Twitter uses


jb55:
looks like this commit branches from an earlier version of the codebase and does not apply cleanly to master. please rebase! :)

jcarucci27:
Fair enough @jb55 I rebased and submitted a new PR

PR #31, opened 2022-09-04
Rebase and redo replies filter to address #23

jcarucci27:
Addresses #23


jb55:
will test this in a bit

PR #44, opened 2022-12-17
Resolved Disappearing Profile Button

sdub18:
If a user has not chosen a profile picture (me), the `ProfilePicView` does not generate a `roboHash` image that can be clicked on to view your on profile. I am not sure if this is a robohash thing, or potentially the return type of `ProfilePicView` that is interfering. But regardless I wanted to put a default SF Symbol there so that users can still click and view their own profile.


jb55:
Looks good! thanks! Merged in d1323ea15c19b0237e7f685e0c646b2164ac9a61

PR #52, opened 2022-12-18
Profile View Modifications

sdub18:
@jb55 Made some modifications to the Profile View. Heavy inspiration from a certain bird app. I also moved the user's public key into a button that copies the public key to the user's clipboard when they tap it. Definitely let me know if you want me to change anything.

Screenshot 2022-12-17 at 11 56 00 PM

### DIFF
--- a/damus/Views/EventView.swift
+++ b/damus/Views/EventView.swift
@@ -93,9 +93,11 @@ struct EventView: View {
                     }
                     .buttonStyle(PlainButtonStyle())
                     TextEvent(inner_ev, pubkey: inner_ev.pubkey)
+                        .padding([.top], 2)
                 }
             } else {
                 TextEvent(event, pubkey: pubkey)
+                    .padding([.top], 6)
             }
         }
     }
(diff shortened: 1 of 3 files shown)


jb55:
this is amazing

jb55:
Thanks! I've cherry picked these on master (ea8394e7cf7ee550529bdfff9ec43e221a79f524) feel free to make any organizing/refactoring code changes you would like.

greimela:
This looks great! 
The only thing I would change is the size of the icons in the round buttons. They could be a bit smaller.

sdub18:
Yeah totally agree @greimela. I was nit-picking that for forever. I'll take another look at it in a future PR

ismyhc:
@sdub18 I noticed that dark mode isn't handled correctly with the new icons. Need to either use a system color that has dark mode ability or create a color set that  is dark mode capable. I would suggest using system colors. 

image

sdub18:
Good catch. Must have missed that during testing. Just put out a PR that fixes it. Thanks!

PR #55, opened 2022-12-19
add dropshadow to profile pic view and realigned event action bar

codycc:
I realigned the buttons on the action bar for a nicer UI. Tested if the double arrow button was not there and they realign for two buttons just as nicely. Also added a dropshadow for the profile picture which seems to be subtle but looks good on light and dark mode.


jb55:
looks a bit weird on dark mode, it's a bright white shadow?

codycc:
Yeah good point, I tweaked it a bit so it's more subtle . but no worries if its not what you're looking for.

BenGWeeks:
We're actually going through a bit of a re-design at the moment, so I think this can be closed for now.

PR #56, opened 2022-12-19
App View Model + Loading Animation + Profile View Dark Mode Fix + Some other small things

sdub18:
Kind of a bigger PR (sorry about that, will keep them smaller and more isolated in the future). One thing I wanted to do was extract a lot of the logic inside of content view and move that inside of its own viewModel so it could be injected into child views using an environmentObject. It makes for a cleaner logic system as we don't really need to pass data around between views, it sorta just exists which is cool.

The other logicy thing I did was make a swipe-down-to-reload extension. I used the `subscribe` function to request a new pull of data from the relays, but please let me know if there's a better way to do that. I wasn't sure.

Other than that, everything in here is very UI-focused. Some improvements to the text layout. Fixing that dark mode bug I accidentally caused, and also adding a new loading view (this uses Apples `redacted` view in combination with a really small shimmer package I found to make it look like its loading. Looks neat but a bit clunky in dark mode as well. Might extract the package out in a future PR so I can tweak it.

I really hope I am not overstepping with all of these changes. Please please please let me know if I am. You have made such an awesome app and personally I think it's criminal for you to have to work on it all by yourself. Especially as Nostr grows every day. If there is anything you want me to change let me know!

Screenshot 2022-12-18 at 10 09 53 PM

## Before
IMG_0408 Medium

## After
IMG_0409 Medium

## Dark Mode
IMG_0410 Medium


jb55:
this is incredible work. this kind of stuff will be rewarded by the nostr fund. thank you!

jb55:
I will test and review soon

jb55:
Looking good! A few comments:

I find the subheadline text everywhere a bit small :(

The shimmering affect in dark mode looks a bit weird

The loading animation doesn't seem to stop until you touch it on iOS

On macos on first load the Posts & Replies tab never shows anything until you switch to Posts

Otherwise looks amazing!

sdub18:
Addressing your PR comments:
- Yeah I can fair. I can see that. I was going for the same formatting of the blue bird, but maybe I just revert those back to body. Or I could also provide some sort of toggle when you first setup the app to choose the font how you would prefer? That might be a little much though. Maybe ill just revert back

- Yeahhhh totally agree. I think I'll just pull out the package and modify so it works well with both.

- Thats a weird one because I don't believe I changed any logic, just the view showing. But ill take a look at this and see if I figure out what's going on.

- I noticed this too. Didn't know if this was me or something else. Will investigate.

Thanks for the comments. I definitely agree with them. I'll take a look at them and see what's going on :)

sdub18:
@jb55 Ready for re-review!

jb55:
ok will try to merge this again

jb55:
perfect, I have a few changes to make but will merge soon

sdub18:
Sick! Sounds great.

jb55:
only had a small tweak to unsubscribe after a pull-down refresh. Still want to make an update so that it doesn't reload when you switch to the global view every time but can make that a follow-up.

Merged in 37460f933b2d0505322de8c27593d28a8c36f463

PR #64, opened 2022-12-20
Update TODO

githubbbie:
next step is to probably remove this file and/or link to the issue tracker page :)


jb55:
we should just remove this file

hewigovens:
Yeah, better to use GitHub project to manage it

PR #68, opened 2022-12-20
animate images everywhere

futurepaul:
https://user-images.githubusercontent.com/543668/208597862-04333cd0-3707-4561-9d81-c94f0118c204.mov

not sure if damus is anti-animation but thought I'd give it a shot. only animates when open in the pop-over view, not in the timeline

I borrowed the cache image and preload settings from the kingfisher demo code:
https://github.com/sean-a-wilson/Kingfisher/blob/1897586f01041d6ddc025714be156aee2d79509d/Demo/Demo/Kingfisher-Demo/SwiftUIViews/AnimatedImageDemo.swift#L41

Not sure why this doesn't take the `resizable()` property but seems to size just fine

also checked on non-animated pngs / jpgs and they still work the same

### DIFF
--- a/damus/Components/ImageCarousel.swift
+++ b/damus/Components/ImageCarousel.swift
@@ -17,11 +17,14 @@ struct ImageViewer: View {
                 VStack{
                     Text(url.lastPathComponent)
                     
-                    KFImage(url)
+                    KFAnimatedImage(url)
+                        .configure { view in
+                            view.framePreloadCount = 3
+                        }
+                        .cacheOriginalImage()
                         .loadDiskFileSynchronously()
                         .scaleFactor(UIScreen.main.scale)
                         .fade(duration: 0.1)
-                        .resizable()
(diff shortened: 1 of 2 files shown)


futurepaul:
https://user-images.githubusercontent.com/543668/208607180-0d4e9fb7-cb0a-409f-95eb-0216baf590ce.mov

jb55:
so good

futurepaul:
had to get rid of `ResizingImageProcessor` to get this to work for the profile pics. not sure what exactly that was doing for us but the help said this:

```
SummaryProcessor for resizing images. If you need to resize a data represented image to a smaller size, use DownsamplingImageProcessor instead, which is more efficient and uses less memory.
```

No idea what all this animation will do to memory usage but excited to find out :)

jb55:
Thanks! merged in 54700e944c03908493d5e843ed3dfd69721ae32c

Issue #50, opened 2022-12-18
Extra space when composing new Note


bebopkid:
When I compose a new note my cursor is added to the end of the helper text (which makes unclear if you need to delete it) and when I start typing there is always a space added before the beginning of my text.

bebopkid:
Fixed in latest release

PR #71, opened 2022-12-20
Move TextEditor cursor to the top left side of the PostView when opening post view 

aki-mizu:
Currently when tapping the plus button to open a post view, the cursor is at the end of the placeholder "Type your post here..." which has the side effect when you tap delete key on keyboard it does not remove the placeholder.
I moved the cursor to the top left side of the PostView to make it feel more nature and refactored the code a bit. 
Please let me know if there's anything missing in my pull request.
Screenshot

### DIFF
--- a/damus/Views/PostView.swift
+++ b/damus/Views/PostView.swift
@@ -15,8 +15,7 @@ enum NostrPostResult {
 let POST_PLACEHOLDER = "Type your post here..."
 
 struct PostView: View {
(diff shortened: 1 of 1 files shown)


jb55:
This is so much better, thank you. Cherry-picked into 90f025564ae25992c693d489ee554e4df1147cf1

aki-mizu:
This fixes #50

PR #74, opened 2022-12-20
add haptic feedback to like action

radixrat:
- add haptic feedback to the like event on EventActionBar View

Not an iOS dev so may be better ways to accomplish this.


jb55:
Cherry-pick into bd5805443447e48b38f28c34bea096b9b327ce4f, thanks!

PR #77, opened 2022-12-20
scoder-1-fix-compilation-errors

scoder1747:
Resolve compilation errors on Xcode 13.4


damus/Views/AddRelayView.swift
```diff
@@ -18,7 +18,7 @@ struct AddRelayView: View {
             Form {
                 Section("Add Relay") {
                     TextField("wss://some.relay.com", text: $relay)
-                        .autocorrectionDisabled(true)
+                        .disableAutocorrection(true)
```
aki-mizu, line 21: It seems disableAutocorrection is deprecated by Apple
https://developer.apple.com/documentation/swiftui/view/disableautocorrection(_:)

PR #80, opened 2022-12-21
Copy image

futurepaul:
about half of this code is just for the sharesheet which tbh doesn't provide a TON of value

ios 16's ShareLink seems a lot better for what we're doing: https://www.hackingwithswift.com/quick-start/swiftui/how-to-let-users-share-content-using-the-system-share-sheet (which I believe also sidesteps any potential permissions issues? like you don't have to add camera roll permissions because it conforms to Transferable protocol (( need to investigate further )))

I'd be happy to rip out the sharesheet stuff or leave it in, up to you

(p.s. not sure why the image always blanks out in the press-hold actions... maybe a Kingfisher bug or setting? it was doing that before I got here though!)
 
Screen Shot 2022-12-20 at 7 56 43 PM

### DIFF
--- a/damus/Components/ImageCarousel.swift
+++ b/damus/Components/ImageCarousel.swift
@@ -47,7 +47,7 @@ struct ImageContextMenuModifier: ViewModifier {
                 Button {
                     UIPasteboard.general.image = someImage
                 } label: {
-                    Label("Copy Image", systemImage: "doc.on.doc")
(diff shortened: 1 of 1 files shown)


jb55:
love it

jb55:
cherry-picked in bb59c2d71c661dc001e0e196bcac4933e52795e1, thanks!

PR #102, opened 2022-12-22
Rounded Images in Timelines

sdub18:
Wanted to make gifs and images a bit more consistent in terms of layout and design. Had some time in between commits and PR reviews for work and wanted to do some Damus coding 😎

Some renderings of what they look like: (Note: I temporarily disabled our sensitive-content filter so I could test on some random images. That filter has been restored before pushing up this change. Just in case seeing photos in the search tab alarmed anyone. Don't worry the dirty photos filter is still there).

Also I kept the full screen view the same.

Screenshot 2022-12-22 at 1 45 10 PM
Screenshot 2022-12-22 at 1 46 24 PM
Screenshot 2022-12-22 at 1 42 20 PM

### DIFF
--- a/damus/Components/ImageCarousel.swift
+++ b/damus/Components/ImageCarousel.swift
@@ -46,21 +46,25 @@ struct ImageCarousel: View {
     var body: some View {
         TabView {
             ForEach(urls, id: \.absoluteString) { url in
-                KFAnimatedImage(url)
-                    .configure { view in
-                        view.framePreloadCount = 3
+                Rectangle()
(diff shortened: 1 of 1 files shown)


jb55:
nice. I'm thinking we can probably remove the link as well for inline images. we just need to add copy image/copy image url

jb55:
I'll merge this in the meantime tho

jb55:
thanks! looks great. merged in 4a83d370d27021ca52222400fa6de0045ab61d3a

PR #105, opened 2022-12-22
Swiftify

thomasRademaker:
I started to make changes that I thought were better Swift coding conventions. But I didn't go too far because then I thought that maybe contributions like this aren't very helpful. 

If you hate this (my feelings won't be hurt) I won't make any more contributions like this. If you like it I can do more of this over time.


thomasRademaker:
Closing because this isn't providing appropriate value to the project.

PR #107, opened 2022-12-22
Add select wallet modal when user pays invoice.

### LINKED ISSUE #79: Add setting for preferred lightning wallet
| example | another example |
| --- | --- |
image | image

Global setting seems nice for this. Default option could be whatever it is right now, and an additional option could potentially be "copy to clipboard"

(I'm planning on working on this just wanted to gather any potential notes here)

suhailsaqan:
Closes #79.

simulator_screenshot_98B93C3B-0123-4B3C-9790-C1618B10956E


jb55:
whoa cool! I use lnlink.app, would it support that?

suhailsaqan:
Yea you could open lnlink.app using `lnlink://` however I'm not sure if it is configured within the app to accept invoices. Most apps use this format: `appname://lightning:invoice`.

jb55:
ok I'll add that to lnlink later... I'll test this in a bit

suhailsaqan:
@jb55 I added an LNLink button to open the app though it won't try to pay any invoice (same thing it does for cash app).

jb55:
testing this now

jb55:
1. it would be good if you could: make one of these your default wallet.

2. it doesn't seem like you can click the pay button in threads anymore after this change

suhailsaqan:
@jb55 How are you thinking the default wallet option is used? Do you mean to add a button that opens the default wallet?

suhailsaqan:
The second problem has always been there and it's because there's an `onTapGesture` on the `View` in which the pay button is contained inside. So it detects both taps (on the `View` and on the button) and thats why it doesn't work properly. I'm trying to fix it but this is my first time working in SwiftUI lol...

jb55:
> I'm trying to fix it but this is my first time working in SwiftUI lol...

yeah I've always struggled with this as well. perhaps we can rework it somehow

suhailsaqan:
> yeah I've always struggled with this as well. perhaps we can rework it somehow

It actually works better in #136 update.

0xtlt:
I sent you a very small PR which contains what makes the "pay" button clickable again.
I imagine we can do better but for a first fix it's good I think

https://github.com/suhailsaqan/damus/pull/2

suhailsaqan:
Thanks!

@jb55 believe this is ready for another review now.

jb55:
kk will try

jb55:
thanks!

PR #109, opened 2022-12-22
Better post editor placeholder

neongreen:
* Smaller padding
* The context menu shows up when you touch the placeholder

Reasoning for smaller padding: it seems to be the default in iOS apps. E.g. Safari search bar:

image

Longform notes in Passwords:

image

Admittedly zero padding seemed even more appropriate but it doesn't mesh well with the letter T.


neongreen:
Btw — sorry for not sending an email patch, but apparently I need to muck around with Google Admin settings to create an app password that can be used to send email. I honestly tried, but failed. Maybe another time.

PR #111, opened 2022-12-22
save the private key to the iOS keychain and not user defaults

thomasRademaker:
User defaults are saved to disk as plain text that can be easily compromised. This PR saves the private key to the iOS keychain which is safe and encrypted.

Vault is a swift package that I created to help save and retrieve items from the keychain. If you'd prefer to use another package or no package at all I can modify this PR.


jb55:
when I tried this on macos, it forced a logout, and when I tried to login it says "invalid key"

thomasRademaker:
I didn't realize this app was building for macOS as well. The latest commit has this fixed.
macOS requires any app using the keychain to include the `Keychain sharing` entitlement. 

Let me know if you see any more issues with this PR.

jb55:
Thanks!

jb55:
noticing a crash on macos when trying to use the privkey... testing

jb55:
oh interesting the string isn't trimmed... weird... fixing.

PR #119, opened 2022-12-23
Loading View Patch

sdub18:
Fixed an issue I noticed with the loading view since we pushed to TestFlight. The `KFAnimatedImage` doesn't really respond to the redacted modifier, so we were getting this weird effect where the Damus logo was showing up instead of a redacted image:

IMG_0420 Medium

I fixed this and also tried to fix the issue with loading. Jack was saying he noticed less notes in his feed and I was worried that had something to do with constantly subscribing every time you refresh. I put some code to unsubscribe first and then resubscribe to see if maybe that might resolve that issue. Let me know what you think @jb55 .

Here is a renderings with the new patch:
Screenshot 2022-12-22 at 5 16 58 PM
(The profile looks light but thats just cuz of the shimmer when I took the screenshot)

### DIFF
--- a/damus/Components/Shimmer.swift
+++ b/damus/Components/Shimmer.swift
@@ -41,6 +41,7 @@ struct ShimmeringView<Content: View>: View {
         _startPoint = .init(wrappedValue: configuration.initialLocation.start)
         _endPoint = .init(wrappedValue: configuration.initialLocation.end)
     }
+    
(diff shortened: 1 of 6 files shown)


jb55:
changes look good! cherry-picked in 549cbb9bce76443aedaec9394f3926acd546667e

Issue #124, opened 2022-12-23
Regarding privacy concerns on loading remote images


radixrat:
Some have raised concerns about images loading from remote servers and logging their IP, I am thinking Damus could provide a toggle that allows users to default to robohash if they don't want to load external profile images if we did another boolean setting check for this setting here in `ProfilePicView`:

```
// TODO: this should be optional setting
let pic = picture ?? profiles.lookup(id: pubkey)?.picture ?? robohash(pubkey)
```

Found some areas of code using `UserDefaults` so think that is a way we can persist these config settings in the app environment.

radixrat:
above PR will add optional toggle under `ConfigView` that allows user to restrict the profile retrieval to robohash at least to potentially address, any feedback welcome (again not iOS developer so having to look up references)

radixrat:
Expanded this option with help from @0xtlt to not just impact profile images, but also inline images in events, prior to this we were choosing to block any images from people that were not friends, now this will give user the option to choose while defaulting to friends of friends.  As described in PR it provides 4 policy levels:

- Everyone
- Friends of Friends
- Friends Only
- Block Images

Updating name of issue to reflect this slight change in direction

radixrat:
Closing as Pr was merged that should handle base case this was intended for.

PR #127, opened 2022-12-23
add optional toggle to load remote images

radixrat:
Potential solution for https://github.com/damus-io/damus/issues/124 once toggled all images come from robohash instead of  loading from remote profiles

### DIFF
--- a/damus/Views/ConfigView.swift
+++ b/damus/Views/ConfigView.swift
@@ -8,6 +8,23 @@ import AVFoundation
 import Kingfisher
 import SwiftUI
 
+enum RemoteImagePolicy: String, CaseIterable {
+    case everyone
+    case friendsOnly
+    case restricted
+}
+
+func remoteImagePolicyText(_ fs: RemoteImagePolicy) -> String {
+    switch fs {
+    case .everyone:
+        return "Everyone"
+    case .friendsOnly:
+        return "Friends Only"
+    case .restricted:
+        return "Restricted (no remote image)"
+    }
+}
+
 struct ConfigView: View {
     let state: DamusState
     @Environment(\.dismiss) var dismiss
@@ -19,8 +36,8 @@ struct ConfigView: View {
     @State var privkey_copied: Bool = false
     @State var pubkey_copied: Bool = false
     @State var delete_text: String = ""
-    @ObservedObject var settings: UserSettingsStore
-    @AppStorage("remote_image_policy") var remote_image_policy: String = "everyone"
+    @EnvironmentObject var user_settings: UserSettingsStore
+    @AppStorage("remote_image_policy") var remote_image_policy: RemoteImagePolicy = .everyone
     
     let generator = UIImpactFeedbackGenerator(style: .light)
     
@@ -133,22 +150,22 @@ struct ConfigView: View {
                 Section("Profile Image Loading Policy") {
                     Menu {
                         Button {
-                            UserDefaults.standard.set("everyone", forKey:"remote_image_policy")
+                            self.remote_image_policy = .everyone
                         } label: {
-                            Text("Everyone")
+                            Text(remoteImagePolicyText(.everyone))
                         }
                         Button {
-                            UserDefaults.standard.set("friends", forKey:"remote_image_policy")
+                            self.remote_image_policy = .friendsOnly
                         } label: {
-                            Text("Friends Only")
+                            Text(remoteImagePolicyText(.friendsOnly))
                         }
                         Button {
-                            UserDefaults.standard.set("restricted", forKey:"remote_image_policy")
+                            self.remote_image_policy = .restricted
                         } label: {
-                            Text("Restricted")
+                            Text(remoteImagePolicyText(.restricted))
                         }
                     } label: {
-                        Text("\(convertImageLoadPolicyTxt(policy:remote_image_policy))")
+                        Text("\(remoteImagePolicyText(remote_image_policy))")
                             .frame(maxWidth: .infinity, alignment: .leading)
                     }
                 }
@@ -196,17 +213,6 @@ struct ConfigView: View {
         .onReceive(handle_notify(.switched_timeline)) { _ in
             dismiss()
         }
-    }
-}
-
-func convertImageLoadPolicyTxt(policy: String) -> String {
-    switch policy {
-    case "restricted":
-        return "Restricted"
-    case "friends":
-        return "Friends Only"
-    default:
-        return "Everyone"
     }
 }
 
--- a/damus/Views/ConfigView.swift
+++ b/damus/Views/ConfigView.swift
@@ -19,8 +19,8 @@ struct ConfigView: View {
     @State var privkey_copied: Bool = false
     @State var pubkey_copied: Bool = false
     @State var delete_text: String = ""
-    
     @ObservedObject var settings: UserSettingsStore
+    @AppStorage("remote_image_policy") var remote_image_policy: String = "everyone"
     
     let generator = UIImpactFeedbackGenerator(style: .light)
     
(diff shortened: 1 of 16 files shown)


radixrat:
Looks like this only restricts for main timeline need to restrict in full profile view as well, may look later

radixrat:
option-off
option-off-view


option-on
option-on-view
option-on-full-profile

radixrat:
when swapping the setting any profiles that were immediately loaded on home view before toggling will still show with previous setting until you scroll down to get more profiles, but I believe this should be blocking anymore remote requests aside from robohash

0xtlt:
Maybe we can have a picker with 3 or 4 options:

- Load all images (insecure) -> will load http and https images
- Load secured images -> will only load https images
- Load contact images -> will only load people in our contact list
- No external images -> will only load robot hash

We can simplify it with 3 options I think

radixrat:
> Maybe we can have a picker with 3 or 4 options:
> 
> * Load all images (insecure) -> will load http and https images
> * Load secured images -> will only load https images
> * Load contact images -> will only load people in our contact list
> * No external images -> will only load robot hash
> 
> We can simplify it with 3 options I think

yeh I think your load all images, load contacts only, and load no external would be good. http/https won’t matter as much if they are worried about ip logging. I’ll have to see if we have a way of determining relation of profile when this image url is set. Thanks.

radixrat:
Added three levels of choice:

- Everyone (default)
- Friends Only
- Restricted 

Screenshot 2022-12-23 at 10 15 11 PM


Screenshot 2022-12-23 at 10 15 18 PM

Not following jb55:

Screenshot 2022-12-23 at 10 15 49 PM

After following:

Screenshot 2022-12-23 at 10 16 14 PM

Everyone else in feed I am not following from this dummy account:
Screenshot 2022-12-23 at 10 16 37 PM

radixrat:
@jb55 let me know if this is something we might be interested in for future or if you have any alternate suggestions, I just heard some people talking about it in forums so thought I would take a shot at it.

0xtlt:
Working for me but "Restricted" should be more expressive for non tech people

damus/Views/ConfigView.swift
```diff
@@ -17,6 +17,7 @@ struct ConfigView: View {
     @State var privkey: String
     @State var privkey_copied: Bool = false
     @State var pubkey_copied: Bool = false
+    @AppStorage("remote_image_policy") var remote_image_policy: String = "everyone"
```
0xtlt, line 20: Don't know if it's possible but can we use enum for the value ?

0xtlt (changes requested):

0xtlt:
Hey!
I just made a PR to your PR branch: https://github.com/radixrat/damus/pull/1

It contains the working version with enum update :)

radixrat:
> Hey! I just made a PR to your PR branch: [radixrat#1](https://github.com/radixrat/damus/pull/1)
> 
> It contains the working version with enum update :)

nice, this is much cleaner, I merged it, thanks!  What is your opinion on a better word for "Restricted"? would "Block All" give a better indication to end user?

0xtlt:
In the code you merged, I changed by "Restricted (no remote image)"

Maybe "Block images" can be better

0xtlt:
Another thing : If we're going to block profile images, we might block post images/gif as well
So, all external contents

radixrat:
> Another thing : If we're going to block profile images, we might block post images/gif as well
> So, all external contents

Yeh believe this could be expanded to cover all image cases, I know there is a block of code that currently restricts image loading to only friends, I think this setting could be leveraged there.

One thing is I think we would want to have a friends of friends option in this case as well

radixrat:
Applied same policy to `should_show_images` we used for profile restrictions, so image loading should follow this setting now as well. One thing I think would be safer is to set the `default` to Friends of Friends instead of Everyone, the reason the code was initially put in place was to stop the porn bots so setting default to everyone would bring that issue back. 

when set to everyone can view images:
Screenshot 2022-12-24 at 9 55 01 AM

When restricted it will block:
Screenshot 2022-12-24 at 9 55 23 AM

0xtlt:
Agreed with the default to friends only

I think we can activate friends of friends but if we add this option, I think we need to keep the friend only option too
And the option friends of friends can be time consuming to develop

Then maybe an improvement for a next time in another PR (or maybe someone can make it)

radixrat:
> Agreed with the default to friends only
> 
> I think we can activate friends of friends but if we add this option, I think we need to keep the friend only option too And the option friends of friends can be time consuming to develop
> 
> Then maybe an improvement for a next time in another PR (or maybe someone can make it)

Friends of friends is already implemented in contacts module, I added it alongside friends so options now are:

- Everyone
- Friends
- Friend of Friends
- Block Images 

think that should cover everything so far.

0xtlt:
Can we swift friends and friends of friends in the list ?

radixrat:
> Can we invert friends and friends of friends in the list ?

done. makes mores sense with this ordering less restrictive -> more restrictive

radixrat:
```
let remote_image_policy: RemoteImagePolicy = RemoteImagePolicy(rawValue: UserDefaults.standard.string(forKey: "remote_image_policy")!) ?? .friendsOfFriends
```
this can cause crash if optional unwrapped is nil, looking at how to clear this

0xtlt (approved):

jb55:
sorry this got left behind... haven't had a change to review this one yet and looks like it needs rebased

radixrat:
rebased to current upstream

BenGWeeks:
I'm actually working on a Core ML porn detector that uses AI to return a value between 0 and 1 for images, where 1 would be extreme content. In the settings you could turn AI detection on.

0xtlt:
Oh!

radixrat:
> Related - I'm actually working on a Core ML porn detector that uses AI to return a value between 0 and 1 for images, where 1 would be extreme content. In the settings you could turn AI detection on.

One issue is if it waits till it’s client side to detect, the image has already been retrieved so if you are at work or something it won’t save you from explaining why porn photos were downloaded 😂

radixrat:
Looks like out of date again I’ll try to rebase, think I need to factor in this new preview pane, if the goal is to not load the images at all from remote sources, I noticed it paints an outline before user clicks so think the image is getting loaded remotely anyways even if show images bool is false.

paulrollo:
Link would be downloaded but not the image itself until actually displayed
or prefetched by the client.

radixrat:
> Link would be downloaded but not the image itself until actually displayed or prefetched by the client.
> […](#)

so the porn ML detection is based on url name instead of actual content of the image?

radixrat:
> > > Link would be downloaded but not the image itself until actually displayed or prefetched by the client.
> > 
> > 
> > > […](#)
> > 
> > 
> > so the porn ML detection is based on url name instead of actual content of the image?
> 
> I haven't submitted the PR for that yet. I don't think it replaces these options, just an additional option (might put it under "experimental"). You pass the image into a method.

Yeh so this is a similar problem I am currently facing with the nice blurred preview pane, the image has to be client side before it can get a blurred version for the ImageCarousal, but by having the image client side to analyze it you've already loaded it from the remote source which means your work network sees you downloaded the porn content.

jb55:
not sure what the state of this is, but ideally I would like to not load blurred images by default. right now they are downloaded and blurred which is not good

radixrat:
> not sure what the state of this is, but ideally I would like to not load blurred images by default. right now they are downloaded and blurred which is not good

I’ll see if I can remove the image preview and add back as links (in case user wants to click through) if their remote load images policy doesn’t align tonight, looks like PR is out of date again so I’ll rebase it

radixrat:
@jb55 @0xtlt okay updated code to only set image URLs into the `img_urls` list of the `render_note_content` if show_images is set to true (as well as the extension matching which was already there), this is based on the remote image policy a user has selected under settings: everyone, friend of friend, friends only, and block images.

Block Images example (pfp is robohash, image is a link):
Screenshot 2023-01-10 at 8 30 55 PM
Screenshot 2023-01-10 at 8 30 25 PM

Everyone image policy display example (pfp renders, images render):
Screenshot 2023-01-10 at 8 31 03 PM
Screenshot 2023-01-10 at 8 31 18 PM
 

Default would be friend of friend, but this should let users set their setting to Block Images and not have to load any note or pfp images from anyone (if at work or something) and then set it back to friend of friend while at home.  People still should use VPN for now, but this might help a bit at least (some people want it for privacy as well (ip tracking) etc).

0xtlt:
Seems good @radixrat 

I'm working btw on an open source projet to delegate the loading of remote contents like videos, gif, images and urls.

This is a first step and maybe my project, the second one to ensure the security of everyone using it :)

radixrat:
@jb55 PR for client configurable image loading is here and ive kept it up to date if you want to try it out.

jb55:
thanks, I merged but made the default everyone like before

PR #132, opened 2022-12-24
Add Wallet Picker Sheet

### LINKED ISSUE #79: Add setting for preferred lightning wallet

leesalminen:
This PR is intended to resolve #79 .

iOS is lame and doesn't let users pick which app they want to open when clicking on a `lightning:` scheme'd URL. They just pick the most recently installed app and shove it down your throat. Gross.

So, we need to add a wallet picker in Damus to give users the power to decide how to send their precious sats. 

This PR adds this ability to Damus.

Screen Shot 2022-12-24 at 4 13 57 PM

The PR, as of now, stands here:

When you click on the lightning icon from a profile this SwiftUI `Sheet` pops up and lets you pick from various well known LN capable wallets, trigger the default OS wallet, or copy the URL to the clipboard. Then the `Sheet` goes away. That's it.

The only thing I haven't implemented yet is the custom URL scheme for each individual app. I just have to do some testing, or find the source code for some other app that currently implements a similar screen.


jb55:
btw we have https://github.com/damus-io/damus/pull/107 open already which I plan to merge soon. sorry :(

jb55:
that one has some issues atm so I'll test this one as well

leesalminen:
OK, cool. No worries. I've just pushed up the final thing to make it all work. Should be ready to test.

jb55:
doesn't seem to work when clicking pay on invoices

jb55:
went with the other pr instead, sorry :[

PR #136, opened 2022-12-25
Improved visual composition for threads

0xtlt:
Hi everyone,

I modified the threads to make them more like those on Twitter.
The functionality is simple: parent events are displayed above, and the direct child events of the currently viewed event are displayed (no other replies).

TODO:

- [x] Home button dismiss the thread view
- [x] Navigation link on others events in the thread view
- [x] Center current event when loading parents
- [x] Extra spaces cleanup
- [x] Subscription rejected: Too many subscriptions: Number of subscriptions must be less than or equal to 10

### DIFF
--- a/damus.xcodeproj/project.pbxproj
+++ b/damus.xcodeproj/project.pbxproj
@@ -130,6 +130,7 @@
 		4CEE2AF9280B2EAC00AB5EEF /* PowView.swift in Sources */ = {isa = PBXBuildFile; fileRef = 4CEE2AF8280B2EAC00AB5EEF /* PowView.swift */; };
 		4CEE2B02280B39E800AB5EEF /* EventActionBar.swift in Sources */ = {isa = PBXBuildFile; fileRef = 4CEE2B01280B39E800AB5EEF /* EventActionBar.swift */; };
 		E990020F2955F837003BBC5A /* EditMetadataView.swift in Sources */ = {isa = PBXBuildFile; fileRef = E990020E2955F837003BBC5A /* EditMetadataView.swift */; };
+		E9E4ED0B295867B900DD7078 /* ThreadV2View.swift in Sources */ = {isa = PBXBuildFile; fileRef = E9E4ED0A295867B900DD7078 /* ThreadV2View.swift */; };
 /* End PBXBuildFile section */
 
 /* Begin PBXContainerItemProxy section */
@@ -308,6 +309,7 @@
 		4CEE2AF8280B2EAC00AB5EEF /* PowView.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = PowView.swift; sourceTree = "<group>"; };
 		4CEE2B01280B39E800AB5EEF /* EventActionBar.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = EventActionBar.swift; sourceTree = "<group>"; };
 		E990020E2955F837003BBC5A /* EditMetadataView.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = EditMetadataView.swift; sourceTree = "<group>"; };
+		E9E4ED0A295867B900DD7078 /* ThreadV2View.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = ThreadV2View.swift; sourceTree = "<group>"; };
 /* End PBXFileReference section */
 
 /* Begin PBXFrameworksBuildPhase section */
@@ -476,6 +478,7 @@
 				4C216F33286F5ACD00040376 /* DMView.swift */,
 				4C06670028FC7C5900038D2A /* RelayView.swift */,
 				E990020E2955F837003BBC5A /* EditMetadataView.swift */,
+				E9E4ED0A295867B900DD7078 /* ThreadV2View.swift */,
 			);
 			path = Views;
 			sourceTree = "<group>";
@@ -788,6 +791,7 @@
 				4C363A9A28283854006E126D /* Reply.swift in Sources */,
 				4C90BD18283A9EE5008EE7EF /* LoginView.swift in Sources */,
 				4C3EA66828FF5F9900C48A62 /* hex.c in Sources */,
+				E9E4ED0B295867B900DD7078 /* ThreadV2View.swift in Sources */,
 				4C3BEFDC281DCE6100B3DE84 /* Liked.swift in Sources */,
 				4C75EFB128049D510006080F /* NostrResponse.swift in Sources */,
 				4CEE2AF7280B2DEA00AB5EEF /* ProfileName.swift in Sources */,
--- a/damus.xcodeproj/project.xcworkspace/xcshareddata/swiftpm/Package.resolved
+++ b/damus.xcodeproj/project.xcworkspace/xcshareddata/swiftpm/Package.resolved
@@ -40,8 +40,8 @@
       "kind" : "remoteSourceControl",
       "location" : "https://github.com/SparrowTek/Vault",
       "state" : {
-        "revision" : "f5707fac23f4a17b3e5ed32dd444f502773615ae",
-        "version" : "1.0.2"
+        "revision" : "87db56c3c8b6421c65b0745f73e08b0dc56f79d4",
+        "version" : "1.0.3"
       }
     }
   ],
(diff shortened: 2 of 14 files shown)


0xtlt:
Here is a quick overview

Simulator Screen Shot - iPhone 14 Pro - 2022-12-25 at 17 36 10

0xtlt:
Now

Simulator Screen Shot - iPhone 14 Pro - 2022-12-25 at 22 40 15

0xtlt:
Update:
Simulator Screen Shot - iPhone 14 Pro - 2022-12-26 at 11 26 33

0xtlt:
This version works, but I still have to retrieve the parent events of the parent events, and then we can test a first stable version :)

jb55:
testing now

0xtlt:
> testing now

I still have some cleanup to do like the extra spaces in the display :)

0xtlt:
Something that I haven't done either and therefore haven't tested is centering the navigation on the current_event.
When the parents have all loaded, it may move the view off the screen

0xtlt:
J'ai rajouté une todo dans le premier commentaire

0xtlt:
Last bug: When I open and close a thread multiples times,

`Subscription rejected: Too many subscriptions: Number of subscriptions must be less than or equal to 10`

0xtlt:
Ok there may be bugs and omissions that I did not see, do not hesitate to send me your feedback

0xtlt:
How could we treat the events that are mentioned in the events?

Imagine we have the following parent event tags:
```json
[
["e", "e1...."],
["e", "e2...."],
["e", "e3...."],
["e", "e4...."],
["e", "e5...."]
]
```

Imagine the content of the current event is: `Hello world! #[0] & #[1]`
Are `#[0]` and `#[1]` supposed to be **e1** and **e2** or **e5** and **e4**?
And which events are considered "parents" of this event? e1, e2, e3 or e3, e4, e5?"

olheadinternet:
#[0] & #[1] = e1 e2

as a user, I want to control the way it's listed, for my playlists I want the reader to go to the top of the thread, as if they were reading a reply thread.

I can also think of an instance where I would want to highlight a note and a *certain* reply, I would also naturally link #[0] as the note and #[2] as the reply... also resulting in them being e1, e2 

could you briefly explain parent?

image

0xtlt:
Okay, According to NIP10:

> ## Marked "e" tags (PREFERRED)
> `["e", <event-id>, <relay-url>, <marker>]`  
> 	
> Where:
> 
>  * `<event-id>` is the id of the event being referenced.
>  * `<relay-url>` is the URL of a recommended relay associated with the reference.  It is NOT optional.
>  * `<marker>` is optional and if present is one of `"reply"` or `"root"`
> 
> **The order of marked "e" tags is not relevant.**  Those marked with `"reply"` denote the `<reply-id>`.  Those marked with `"root"` denote the root id of the reply thread.
> 
> >This scheme is preferred because it allows events to mention others without confusing them with `<reply-id>` or `<root-id>`.  

Then, mentions are based on non reply or root marker

```json
{
...
tags: [
  ["e", "e1....", "wss...", "root"],
  ["e", "e2....", "wss...", "reply"],
  ["e", "e3....", "wss...", "reply"],
  ["e", "e4....", "wss...", "reply"],
  ["e", "e5....", "wss..."],
  ["e", "e6....", "wss...", "reply"],
  ["e", "e7....", "wss..."],
]
}
```

Here the `#[0]` is **e5** according to NIP10 and `#[1]` is **e7**

0xtlt:
I will implement this

0xtlt:
@jb55 what do you think ?

0xtlt:
And if a e tag is not used in the content, I will consider it as "reply".

it will not break the current threads and at the same time it will be necessary to update the reply function of damus

0xtlt:
Or directly use the index, I will take the following event for the exemple

```json
{
  "pubkey": "5fd693e61a7969ecf5c11dbf5ce20aedac1cea71721755b037955994bf6061bb",
  "content": "instead of reading that poorly worded explanation it’s probably easier to just look at this note #[3]",
  "id": "d8c24ac1e4f7f64c8b85413f3f2ba3031b5d8a01c5ca662b7a72ecf935933ef6",
  "created_at": 1672127500,
  "sig": "e1b4f8e158d7626d102b6b90c93b7e4fef8a3a6542bf96bcfc0e1406cdf3b7d1d085820a4ba7488361f03c0497791fff75806b3435536609bd8abe0b90cb0d2d",
  "kind": 1,
  "tags": [
    [
      "e",
      "15cf603a7ebaf94d30e59e92b71690fcd97f44013c57eafca9d73fd824c3c98e"
    ],
    [
      "e",
      "5830ce3b3420c6559ddd715e2570890abe534357981973b8f9bb0b5fd4ffafa2"
    ],
    [
      "p",
      "884704bd421721e292edbff42eb77547fe115c6ff9825b08fc366be4cd69e9f6"
    ],
    [
      "e",
      "e6267b81778299576ad0655b6d77ebd24c1206103b46d4451c35871bbaed1619"
    ]
  ]
}
```

0xtlt:
Ok now it's fixed with the following logic: If the index of the tag is found in the event content in the format: `#[index]`, then, skip it as parents

0xtlt:
Simulator Screen Shot - iPhone 14 Pro - 2022-12-27 at 13 48 45

nvk:
what does it look like if this note has a reply? 
image

0xtlt:
> what does it look like if this note has a reply?
> image

The preview only displays the "direct" content of the event, it will not load the responses or the "parents"

nvk:
Want to try to preview (or partial) a hop on each direction on the same thread?

0xtlt:
Have a little bit of the thread messages above and below the preview? @nvk

olheadinternet:
are those note links even clickable? and same question for the notes they lead too, are the note links on these subnotes there clickable. in 0.1.8 (4) the second tier of note links break.

nvk:
> Have a little bit of the thread messages above and below the preview? @nvk

Now that I started drawing a quick vis, will be too cluttered. So ignore my last. Once we have reply counts it will be a bit less confusing.

now, I do think anytime we have note’s hex link, we should put the preview of it if the relay has that data.

0xtlt:
> > Have a little bit of the thread messages above and below the preview? @nvk
> 
> Now that I started drawing a quick vis, will be too cluttered. So ignore my last. Once we have reply counts it will be a bit less confusing.
> 
> now, I do think anytime we have note’s hex link, we should put the preview of it if the relay has that data.

I prefer to limit the number of previews in a single event view.

To avoid having displays that do not make sense.

What do you think

0xtlt:
> are those note links even clickable? and same question for the notes they lead too, are the note links on these subnotes there clickable. in 0.1.8 (4) the second tier of note links break when in thread/chat.

Yes they are, but because of the display system used to show the "links", I think it's not possible to prevent the navigation from breaking without re-code this part

Actually this PR is already big and I prefer to keep this "bug" for another PR

nvk:
> > > Have a little bit of the thread messages above and below the preview? @nvk
> > 
> > 
> > Now that I started drawing a quick vis, will be too cluttered. So ignore my last. Once we have reply counts it will be a bit less confusing.
> > now, I do think anytime we have note’s hex link, we should put the preview of it if the relay has that data.
> 
> I prefer to limit the number of previews in a single event view.
> 
> To avoid having displays that do not make sense.
> 
> What do you think

That’s fine approach. But hex is very confusing machine content. So maybe replace the hex with a “note date” ?

0xtlt:
> note date

I agree, on the other hand I will surely do it in another PR to avoid having a merge that hurts haha

0xtlt:
this can be done at the same time as a redesign of the textual content of the messages as I sait below

> > are those note links even clickable? and same question for the notes they lead too, are the note links on these subnotes there clickable. in 0.1.8 (4) the second tier of note links break when in thread/chat.
> 
> Yes they are, but because of the display system used to show the "links", I think it's not possible to prevent the navigation from breaking without re-code this part
> 
> Actually this PR is already big and I prefer to keep this "bug" for another PR because it's not caused by this PR and is already in the main
> 
> Another think I can do here is to prevent navigation in preview links which is actually a bug

jb55:
sorry will test this soon... was setting up the new relay yesterday.

0xtlt:
Take your time @jb55, it’s already great to have launched the project and it’s the holidays!🎄

jb55:
pretty happy with this! I'll make a few tweaks and merge

jb55:
ah it's still a bit buggy and seems to have introduced the navigation popping bug again :(

jb55:
I'm tempted to just merge this though since it's so much better already...

jb55:
merged in d1cbf74, thanks!

PR #143, opened 2022-12-26
fix: Add Divider() to bottom of navigation bar

realprogrammersusevim:
It bugged me whenever I opened the app that the icons in the bottom main tab bar weren't centered and were too close to the bottom of the screen. I added a divider to the bottom of the tab bar to center it and give a bit more space from the bottom of the screen (or window in macOS). In my personal opinion it makes it look much better.


realprogrammersusevim:
Here's a screenshot to demonstrate the issue.

Screenshot 2022-12-26 at 11 51 59 PM

The left side was built with the change while the right side is the old version.

realprogrammersusevim:
This pull request #161 has a better solution. Closing.

Issue #106, opened 2022-12-22
Pinch Zoom Profile Image feature in separate window


scoder1747:
Can we add a feature to see the Zoomed Profile Image in a separate window as below? 
![ezgif com-gif-maker](https://user-images.githubusercontent.com/120697811/209233390-10f86c95-4774-4ff5-97a5-07098636ecd7.gif)

PR #149, opened 2022-12-26
added ability to zoom profile pic on double-tap

niteshbalusu11:
Signed-off-by: Nitesh Balusu <niteshbalusu@icloud.com>

Added ability to zoom photo on double tap, Single tap zooming is adding complexity because the same ProfilePicView is used in a lot of places.

### DIFF
--- a/damus/Views/ProfilePicView.swift
+++ b/damus/Views/ProfilePicView.swift
@@ -34,16 +34,18 @@ func pfp_line_width(_ h: Highlight) -> CGFloat {
 
 struct InnerProfilePicView: View {
     @Environment(\.redactionReasons) private var reasons
-    
+
     let url: URL?
     let pubkey: String
     let size: CGFloat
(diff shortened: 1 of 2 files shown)


niteshbalusu11:
https://user-images.githubusercontent.com/84944042/209568769-e653540e-2c52-4636-bf95-2ed9ca83f1c2.mov

niteshbalusu11:
https://github.com/damus-io/damus/issues/106

jb55:
I feel like this tap gesture shouldn't be on the inner profile pic, why not just add it to the profile view?

niteshbalusu11:
> I feel like this tap gesture shouldn't be on the inner profile pic, why not just add it to the profile view?

Well that was stupid of me, fixed it. Also switched it to single tap now.

jb55:
perfect, thanks!

PR #161, opened 2022-12-27
Tab Bar Padding Issue

sdub18:
Some users were messaging me saying the tab bar is awkward on their device's layout (older phones like the iPhone 7 or SE). This PR just adds a bit of padding on the tab bar to resolve that.

Screenshot 2022-12-27 at 8 16 21 AM


jb55:
🤙🏼

PR #162, opened 2022-12-27
Minor UI Fix for Followers List

famoseagle:
Before:
image

After:
image

### DIFF
--- a/damus/Views/FollowingView.swift
+++ b/damus/Views/FollowingView.swift
@@ -53,8 +53,8 @@ struct FollowersView: View {
                     FollowUserView(target: .pubkey(pk), damus_state: damus_state)
                 }
             }
+            .padding(.horizontal)
         }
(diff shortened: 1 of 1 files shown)


damus/Views/FollowingView.swift
```diff
@@ -52,6 +54,7 @@ struct FollowersView: View {
                 }
             }
         }
+        .padding(.horizontal)
```
OlegAba, line 57: I believe the padding modifier should be attached to the LazyVStack to avoid moving the scroll indicator. In my opinion, it should also be horizontal and vertical. Add all changes to the FollowingView as well.

IMG_4465

famoseagle: Thanks. Fixed.

jb55:
looks good!

jb55:
merged, thx!

PR #163, opened 2022-12-27
Add keyboard shortcuts to switch main views and post

realprogrammersusevim:
I added several keyboard shortcuts so users can switch between the main views of "Home", "DMs", "Search", and "Notifications" with the shortcuts "CMD+1", "CMD+2" etc.. Those shortcuts made the most sense to me but they're definitely up for discussion.

Those were first since they're the easiest to add. I'm going to work on getting
a refresh ("CMD+SHIFT+N" or "CMD+R") shortcut and a "CMD+N" shortcut for a new
post.

### DIFF
--- a/damus/Views/MainTabView.swift
+++ b/damus/Views/MainTabView.swift
@@ -76,14 +76,11 @@ struct TabBar: View {
         VStack {
             Divider()
             HStack {
-                TabButton(timeline: .home, img: "house", selected: $selected, new_events: $new_events, action: action)
-                TabButton(timeline: .dms, img: "bubble.left.and.bubble.right", selected: $selected, new_events: $new_events, action: action)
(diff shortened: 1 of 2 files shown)


realprogrammersusevim:
Adding those actions to the menu bar with the associated keyboard shortcut displayed would also be helpful so users could immediately learn about the shortcuts without having to read documentation or screw around.

realprogrammersusevim:
Those keyboard shortcuts should be good for now. Since Damus loads new posts automatically a refresh shortcut isn't needed. I'll look into creating a better menu bar soon but I think that should be a separate pull request.

I can include a screen recording of the shortcuts working if you want.

jb55:
Trying this now

jb55:
merged, thx!

PR #164, opened 2022-12-27
Nostr Event Attachments UI Adjustments

sdub18:
This PR adds some UI fixes to:
- Blur and opaque non-friend images rather than only display the link (A happy medium for now I think)
- Remove URLs in content text when image is displayed
- Show non-image URLs as clickable link views (more to be done)
- Adjustments to the Pay button on invoices.

Small little adjustments. Want to eventually add metadata for links so they can be a bit more colorful. Still working on that. Let me know what you guys think!

Screenshot 2022-12-27 at 12 41 55 PM
Screenshot 2022-12-27 at 12 41 36 PM
Screenshot 2022-12-27 at 12 40 59 PM

### DIFF
--- a/damus.xcodeproj/project.pbxproj
+++ b/damus.xcodeproj/project.pbxproj
@@ -7,6 +7,7 @@
 	objects = {
 
 /* Begin PBXBuildFile section */
+		3165648B295B70D500C64604 /* LinkView.swift in Sources */ = {isa = PBXBuildFile; fileRef = 3165648A295B70D500C64604 /* LinkView.swift */; };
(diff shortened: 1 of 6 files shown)


jb55:
very nice!

sdub18:
Detailed links are running using async await but are causing some view warnings for a reason I cannot seem to figure out. If someone has time to check this out that would be great. It's good to merge in but ideally would like to remove the warnings before it goes in.

Screenshot 2022-12-27 at 5 41 07 PM

jb55:
I'm pretty happy with all these changes, merged!

PR #171, opened 2022-12-28
Make links clickable in profile view

lionello:
Just a quick thing I thought I'd code up to get familiar with the Damus code. Some folks have URLs in their profile (like @jb55 ) and it would be nice to be able to click those links.


jb55:
thanks!

PR #172, opened 2022-12-28
Proof of work

0xtlt:
Because the application must remain understandable to everyone, it will not be possible to choose its PoW level, instead 4 distinct levels will be available:

- Disabled: No filter and PoW on sent messages
- Level 1: PoW difficulty 5 (may change in the future)
- Level 2: PoW of difficulty 10 (may change in the future)
- Level 3: PoW of difficulty 15 (may change in the future)

It would be necessary to find a "naming" simpler than PoW on the other hand, like "security filter" or "anti spam filter"

----

- The "parent" messages of a message with a proof of work that is validated by the client will by default be validated without a check to avoid breaking the threads

### DIFF
--- a/damus/Views/ConfigView.swift
+++ b/damus/Views/ConfigView.swift
@@ -7,6 +7,13 @@
 import AVFoundation
 import SwiftUI
 
+enum ProofOfWorkPolicy: Int, CaseIterable {
+    case disabled = 0
+    case little = 5
+    case middle = 10
+    case max = 25
+}
+
 struct ConfigView: View {
     let state: DamusState
     @Environment(\.dismiss) var dismiss
--- a/damus/Views/ConfigView.swift
+++ b/damus/Views/ConfigView.swift
@@ -9,10 +9,24 @@ import SwiftUI
 
 enum ProofOfWorkPolicy: Int, CaseIterable {
     case disabled = 0
-    case little = 5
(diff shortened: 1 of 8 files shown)


damus/Views/ConfigView.swift
```diff
@@ -7,6 +7,27 @@
...
 import SwiftUI
 
+enum ProofOfWorkPolicy: Int, CaseIterable {
+    case disabled = 0
+    case level1 = 5
+    case level2 = 10
+    case level3 = 25
+}
+
+func proofOfWorkPolicyText(_ policy: ProofOfWorkPolicy) -> String {
```
lionello, line 17: This could be a member of the enum

0xtlt: Will check tomorrow 👍

jb55:
pow is already implemented in damus (count_hash_leading_zero_bits) it's just not used because it's not that useful for spam prevention at this time. all levels described here are extremely trivial to create on a single-core CPU in under a second.

BenGWeeks:
My opinion is that Sats are the PoW that's already been done :-)

Shall we close this?

jb55:
yeah PoW is a dead end imo

0xtlt:
Yep, I agree

I close this PR :)

PR #181, opened 2022-12-29
Fix issue where boosted post causes duplicate thread

niteshbalusu11:
Signed-off-by: Nitesh Balusu <niteshbalusu@icloud.com>

Fixed the issue where every boost is showing up as part of the thread.


niteshbalusu11:
Before:

Screen Shot 2022-12-29 at 2 14 46 AM


After:

Screen Shot 2022-12-29 at 2 13 52 AM

jb55:
This has already been fixed/ we should just be fetching kind1 instead of doing hacks like this after the fact.

PR #185, opened 2022-12-29
Fix padding on TabBar view

OlegAba:
Ideally, the TabBar should be refactored to use TabView and tabItems so that it follow the safe area properly. This should be sufficient temporary fix for the time being.

**Old Devices**

Before             |  After
Screenshot 2022-12-29 at 2 03 22 PM |  Screenshot 2022-12-29 at 2 07 50 PM

**Modern Devices**

Before             |  After
IMG_4484 |  IMG_4485


jb55:
oops I fixed this already without seeing this PR, closing

PR #191, opened 2022-12-30
Beautify Lightning invoices

BenGWeeks:
Hi JD,

I've applied some on-brand styling to the Lightning invoice block. I've also allowed some space at the bottom of the invoice for the carousel, which looks like it creates a TabView of invoices if multiple invoice blocks exist in the event. However, I couldn't generate a post with multiple invoices, so I wonder if that's been implemented.

The invoices will look like this:

image

Let me know if you want more styling applied as per the Figma designs here:

Password: **keypair**

Cheers,

Ben


BenGWeeks:
Please also see: Beautify tab buttons & new post icon (#196)

BenGWeeks:
Will submit an update in a few days that tweaks `padding(.bottom)` based on whether invoice is part of a carousel or not. Or we move carousel to under the invoices.

Having said that, in my testing I didn't get multiple invoice blocks to appear in a single post?!

schlunsen:
I'm not a big fan of the blue pay button tbh and I think a white would work better. 

Also it could be nice if the spacing/padding in the top and bottom is the same like it is with the sides.

Just my 2 cents

BenGWeeks:
Yeah, me neither, but the `.accentColor` has now been updated, so we could change that to be consistent.

Yeah, we should tweak it so the padding at the bottom only grows if it's in a carousel (which I have never seen them as, to be honest!).

BenGWeeks:
New design for this inbound.

Issue #152, opened 2022-12-26
(enhancement) long press, or swipe left, to copy relay address from relay list


olheadinternet:
I'm sure when relay lists come to profiles this will be solved differently.

dannmat:
I can pick this up now

PR #197, opened 2022-12-31
Added copy relay action

dannmat:
Added #152 to allow user to copy a relay after a long press (in the same context menu as delete relay)

Preview:
simulator_screenshot_DF2C5085-D7B6-4647-970D-74A388F1788A


jb55:
Thanks!

Issue #142, opened 2022-12-26
pfp's get real big when multi-tasking


olheadinternet:
https://i.imgur.com/djWhl9y.mp4

PR #200, opened 2022-12-31
Fix Expanding Profile Picture

joelklabo:
This fixes the issue with the profile picture expanding when the app is going to multitasking (inactive state).

Fixes: #142 

We were checking the `redactionReasons` to determine which profile picture we should show. When you enter the `inactive` state by switching apps etc. The `readactionReasons` are populated so we would fall into the else statement there. I think normally you would show a placeholder or something similar. But, we were just showing the profile picture (non-animated version) and it didn't have any of the sizing modifiers on it. 

If responding to redaction is important we could reintroduce the if/else and show a placeholder there but I don't think it makes sense if we aren't hiding any information.

Profile pictures with fix:
Simulator Screen Recording - iPhone 14 Pro - 2022-12-31 at 14 06 52

### DIFF
--- a/damus/Views/ProfilePicView.swift
+++ b/damus/Views/ProfilePicView.swift
@@ -33,8 +33,6 @@ func pfp_line_width(_ h: Highlight) -> CGFloat {
 }
 
 struct InnerProfilePicView: View {
-    @Environment(\.redactionReasons) private var reasons
-
     let url: URL?
     let pubkey: String
     let size: CGFloat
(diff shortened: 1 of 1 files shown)


jb55:
oh nice

jb55:
thanks! cherry-picked into 759ba398cc022564ca72efad762ee032db808dfb

joelklabo:
Yeah, that took me a while to figure out 😂

PR #214, opened 2023-01-02
Replace deprecated Data.withUnsafeMutableBytes call

tyiu:


damus/Nostr/NostrEvent.swift
```diff
@@ -448,7 +448,7 @@ func hex_encode(_ data: Data) -> String {
 func random_bytes(count: Int) -> Data {
     var data = Data(count: count)
     _ = data.withUnsafeMutableBytes { mutableBytes in
-        SecRandomCopyBytes(kSecRandomDefault, count, mutableBytes)
+        SecRandomCopyBytes(kSecRandomDefault, count, mutableBytes.baseAddress!)
```
tyiu, line 451: Someone should double-check my work. I don't know if this is actually the right thing to do.

jb55:
thanks this was annoying me

PR #215, opened 2023-01-02
Hide Edit Button on Profile Page for the user not logged in w/Private Key

scoder1747:
Hide Save button when not logged in with Private key

Simulator Screen Shot - iPhone 14 Pro - 2023-01-02 at 12 11 55

### DIFF
--- a/damus/Views/EditMetadataView.swift
+++ b/damus/Views/EditMetadataView.swift
@@ -175,11 +175,9 @@ struct EditMetadataView: View {
                     }
                 })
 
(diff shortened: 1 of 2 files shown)


jb55:
perhaps we shouldn't even show the edit button? since you can't edit anyways?

scoder1747:
> perhaps we shouldn't even show the edit button? since you can't edit anyways?

I have addressed your feedback, thank you :)

jb55:
thanks!~

PR #220, opened 2023-01-03
Beautify event actions

BenGWeeks:
Hi all,

OK, did a small one this time that is consistent with the other "Beautify *" PRs (#203, #196 and #191), that apply the following UI:

image

Namely:

- Added on-brand iconography for the event actions (including hidden actions)
- Applied a custom "accentColor"
- Applied gradient style to highlighted actions
- Made use of "Spacer()"

The updates are in line with the designs in Figma that you can see here:

Password: **keypair**

If we merge this one, propose I can work on further enabling the hidden actions ;-)


BenGWeeks:
image

bitcoinuser:
Why this settings icon on top right is so ugly?

BenGWeeks:
In the toolbar top/right?

BenGWeeks:
Propose to address in a future update.

bitcoinuser:
> image
The Settings tip right button here.

BenGWeeks:
Related to #226. I created the thumbnail for the hand in Ps but that image may be better (see "ic-like", using "ic-like-medium.png" and "ic-like-light.png" from memory).

BenGWeeks:
This should be easier to merge than #203, or you need it rebased?

jb55:
I'm not sure I like the look of the hollow purple shaka... should probably be filled?

BenGWeeks:
Oh, "shaka" - is that how you it should be referred?! :-) Yeah, I did wonder about that. #226 might be able to help with that? :-)

elidyweaver:
Adding my two cents. When you have a row of glyphs, they should be as uniform as possible. Size, style, and in this case, if they're outlined, they should all be outlined - unless you're trying to highlight one above the rest. 

A good question to ask would be what's most important to Damus in that row of reactions? That's what should be unique or different, if wanted. Otherwise the difference can come when you click it. 

In that specific case of the "shaka," the issue might have been using a gradient, the darker color got too lost in the background, so you cannot make out the shape as easily. 

Hey, and if any of you want to talk about UI/UX design issues, you're welcome to join the discord: https://discord.gg/bJrWFds6je

BenGWeeks:
Happy to go in direction you think @jb55

BenGWeeks:
Would be good to have your input on this one @jb55 before I make amendments (suggest one of the first 2 of the 3 possibilities above).

BenGWeeks:
Closing as new re-design happening anyway.

PR #223, opened 2023-01-03
Always check `SecRandomCopyBytes` return value

hewigovens:
From https://developer.apple.com/documentation/security/1399291-secrandomcopybytes

> Always test the returned status to make sure that the array has been updated with new, random data before trying to use the values


jb55:
Thanks this looks much better

PR #225, opened 2023-01-03
Fix detection of email addresses in profiles

lionello:
This fixes the logic added in 6f5f86114b0948cac286f2b1594e8f3f831e9032 by supporting email addresses (and any other schemes.) The `NSDataDetector` adds the right scheme for us, but unfortunately it defaults to `http` for URLs without scheme.

Signed-off-by: Lionello Lunesu <lio+git@lunesu.com>


jb55:
cool thx!

PR #226, opened 2023-01-03
Switching from ❤️ to 🤙 images

elidyweaver:
Uploading four variations of "🤙" images:
black outline
white outline
solid black
solid white

In three different sizes:
20x20, @2x, and at @3x

Added contents.json imageset file

*They are only the images, I have not changed any code within the app.

Would you please add the images where they are needed? or point me in the right direction? I have much yet to learn.

I hope these are the versions you need. If there are any adjustments, please let me know.


BenGWeeks:
Nice. Relates to #220. I didn't define 2x / 3x etc though (only 24x24px 1x). Not sure implications of that? Could maybe replace the "ic-like" with your image (I just put something together quite quickly and maybe yours look more like hands).

NB That Pr doesn't use a fill color, it actually uses a LinearGradient of Damus colours when highlighting (see screenshot in PR).

For Light mode I used "DamusMediumGrey" (see Hex in color set) and Dark used "DamusLightGrey" from memory as apposed to just white/black to be consistent with the "Beautify Tabs" PR.

elidyweaver:
Thank you so much for the thoughts, @BenGWeeks. I will be looking through your files, correcting colors, and reuploading.

BenGWeeks:
What are the benefits and uses of 2x and 3x versions? Is that for hard of site users who scale their screens or something?

jb55:
thanks, will update the event action bar

elidyweaver:
> What are the benefits and uses of 2x and 3x versions? Is that for hard of site users who scale their screens or something?
Sorry, I was afk most of yesterday.  Apple recommends it for larger screens (like iPads) and higher resolutions (retina). (Maybe it has some accessibility benefits, I'm not sure, I need to dig deeper into that.)

elidyweaver:
Thank you Damus team (@BenGWeeks, @jb55+) for helping through all the finishing process and making it available in app.

PR #232, opened 2023-01-03
Fix padding on search results view

OlegAba:
Before             |  After
IMG_4536  |  IMG_4537


jb55:
🙏

PR #233, opened 2023-01-03
Fix home view moving after selecting from search result

OlegAba:
Before             |  After
 |


jb55:
Thanks!

PR #235, opened 2023-01-03
Adding a relay. We have been running servers for years for MMO games.…

etemiz:
Adding a relay. We have been running servers for years for MMO games. We think seriously in this relay operation.
Bare metal - dedicated fast server. 12 cores, 64GB, nvme, ...


etemiz:
wss://nostr.mom

jb55:
ok I will consider it if we start to lose any, we have plenty of recommended relays already

PR #238, opened 2023-01-03
Fix ASCII shrug

lionello:
Changelog-Changed: Fix ASCII shrug

### DIFF
--- a/damus/Util/Markdown.swift
+++ b/damus/Util/Markdown.swift
@@ -21,7 +21,7 @@ public struct Markdown {
             .init(interpretedSyntax: .inlineOnlyPreservingWhitespace)
 
         // TODO: escape unintentional markdown
-        var escaped = content.replacingOccurrences(of: "\\_", with: "\\\\\\_")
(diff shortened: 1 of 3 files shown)


lionello:
@jb55 have a look

damus/Util/Markdown.swift
```diff
@@ -20,7 +20,9 @@ public struct Markdown {
         let md_opts: AttributedString.MarkdownParsingOptions =
             .init(interpretedSyntax: .inlineOnlyPreservingWhitespace)
 
-        if let txt = try? AttributedString(markdown: content, options: md_opts) {
+        // TODO: escape unintentional markdown
+        var escaped = content.replacingOccurrences(of: "\\_", with: "\\\\\\_")
```
lionello, line 24: escaping all backslashes is unnecessary (see tests), so we only need to escape the ones where backslash does escaping for a valid markdown char, like `_`.

lionello: Yeah, I'm not too happy with this hack, but when note content can be sometime markdown, sometimes regular text, we need something to handle to corner cases.

Issue #188, opened 2022-12-30
Poor performance with large size GIFs


OlegAba:
Profile pictures that use large file size images (mostly GIFs) cause very poor performance - scrolling lag. Look at the video below for profile @npub1gpppr3hfwcl5njxqmp3zumaly9j4pd6gvapywsq2y5n7p7278fus6nz7l5 who uses a 65MB image. The Kingfisher library has a downsampling modifier. Unfortunately, it does not work for [animated images](https://github.com/onevcat/Kingfisher/issues/1940) . I think the best approach would be to not animate any images over a byte limit (100KB?). 

https://user-images.githubusercontent.com/19398259/210023069-9aa51f74-c484-49bf-8e87-ef9de22a9916.mov

jb55:
Yes we should do that

PR #240, opened 2023-01-04
Add KingFisher downsampler processor for large images

OlegAba:
Ref: [Issue #188](https://github.com/damus-io/damus/issues/188)

Adding a background callback and processing queue significantly improved performance. However, once a large image/GIF was loaded it would still cause scrolling lag. This PR adds a processor to `KFAnimatedImage` that will downsample any image over `maxSize`. I use profile `@npub1gpppr3hfwcl5njxqmp3zumaly9j4pd6gvapywsq2y5n7p7278fus6nz7l5` as an example for the before/after below. If you test this yourself, note that it will take a while to load as the profile picture is ~65MB.

Before             |  After
 |  

### Considerations
Not sure what the best `maxSize` and `downsampleSize` would be. Most GIFs appear to be over my original proposal of 100KB limit. A max size of 1MB seems reasonable for GIFs and did not cause any performance issues during my tests. I suspect some users will complain that their profile pictures are not animating after the change - a UI warning should be added reflecting whatever size is decided upon (`EditMetaDataView`). `PFP_SIZE` could be used for the downsample size, but this will be make the image low resolution when it is expanded. Open to any suggestions.

### DIFF
--- a/damus/Views/ProfilePicView.swift
+++ b/damus/Views/ProfilePicView.swift
@@ -55,6 +55,7 @@ struct InnerProfilePicView: View {
             KFAnimatedImage(url)
                 .callbackQueue(.dispatch(.global(qos: .background)))
                 .processingQueue(.dispatch(.global(qos: .background)))
+                .appendProcessor(LargeImageProcessor())
                 .configure { view in
                     view.framePreloadCount = 1
                 }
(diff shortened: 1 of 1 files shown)


jb55:
I attempted to do this but this seems much better.

PR #244, opened 2023-01-04
Beautify new post button

BenGWeeks:
Minor amendment to the post button (achieved using shapes and text rather than images).

image

FYI It would be good to fix why this isn't appearing in the preview mode of the app.

ChangeLog-Changed: Updated new post button to gradient color (@npub1jutptdc2m8kgjmudtws095qk2tcale0eemvp4j2xnjnl4nh6669slrf04x)

### DIFF
--- a/damus/Assets.xcassets/Colors/DamusBlue.colorset/Contents.json
+++ b/damus/Assets.xcassets/Colors/DamusBlue.colorset/Contents.json
@@ -1,0 +1,38 @@
+{
+  "colors" : [
+    {
+      "color" : {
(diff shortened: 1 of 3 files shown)


jb55:
thanks! I have a variant of this in the merge queue.

BenGWeeks:
No problem.

Is the "Chang**g-Changed" in the right place?

BenGWeeks:
We able to have a catch-up at some point @jb55 on the direction you want to take things and priorities? Happy to help - can just be a lot more productive I think.

PR #247, opened 2023-01-04
Beautify the profile back button and text to white

BenGWeeks:
Updated the text colour to white for the back button and text (top left).

ChangeLog-Changed: Updated new post button to gradient color (@npub1jutptdc2m8kgjmudtws095qk2tcale0eemvp4j2xnjnl4nh6669slrf04x)

### DIFF
--- a/damus/Util/Theme.swift
+++ b/damus/Util/Theme.swift
@@ -9,10 +9,10 @@ import Foundation
 import UIKit
 
 class Theme {
-    static func navigationBarColors(background : UIColor?,
-       titleColor : UIColor? = nil, tintColor : UIColor? = nil ){
+    static func navigationBarColors(background : UIColor?, titleColor : UIColor? = nil, tintColor : UIColor? = nil) {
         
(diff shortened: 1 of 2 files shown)


jb55:
back button not visible on light mode :(

BenGWeeks:
Got a screenshot? Shouldn't this also be overlayed over the ostrich image?

BenGWeeks:
Here's what I see:

image

(Some drop shadow would be good but that alluded me)

BenGWeeks:
Any updates on this one @jb55? You not seeing the same?

jb55:
I'll take another look. now that we have banner images I think this will always be an issues unless we have a better solution

BenGWeeks:
Yeah, would be great to have a drop-shadow to add contrast behind it. I'll look at this again but for whatever reason I couldn't apply a drop-shadow before.

BenGWeeks:
I think this is resolved now anyway.

PR #248, opened 2023-01-04
Make Event in Reply View Scrollable

joelklabo:
Embed event content in a scroll view when replying so it doesn't collapse the UI

Before:
Simulator Screen Recording - iPhone 14 Pro - 2023-01-04 at 14 40 50

After:
Simulator Screen Recording - iPhone 14 Pro - 2023-01-04 at 14 38 25

### DIFF
--- a/damus/Views/ReplyView.swift
+++ b/damus/Views/ReplyView.swift
@@ -33,7 +33,9 @@ struct ReplyView: View {
                     .foregroundColor(.gray)
                     .font(.footnote)
             }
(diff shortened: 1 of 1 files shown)


0xtlt:
Mmmh can we disable the preview and trim the text?

joelklabo:
Yeah, that could work too. But you may want to refer back to some part of a long message to reply too. Probably not common but it's nice to be able to

jb55:
I was thinking about trimming but this works for now

PR #253, opened 2023-01-05
Deeplink launching to profiles and notes

BenGWeeks:
Showing a blank screen at the moment as damus_state not yet initialised I presume, but putting here so you guys knows this is pretty much done.

TO DO:

- Resolve damus_state initialisation [RESOLVED]
- Add instructions to Readme.md for users
- Increased error handing in DeeplinkManager

Will accept commits to the fork! :-)

ChangeLog-Changed: Added deeplink ability (using nostr://io.damus.nostr/npub1jutptdc2m8kgjmudtws095qk2tcale0eemvp4j2xnjnl4nh6669slrf04x format) (@npub1jutptdc2m8kgjmudtws095qk2tcale0eemvp4j2xnjnl4nh6669slrf04x)
ChangeLog-Changed: Added additional error handling to loading contact events (@npub1jutptdc2m8kgjmudtws095qk2tcale0eemvp4j2xnjnl4nh6669slrf04x)


jb55:
pretty sure we handle this already? I'm confused. see onOpenURL in ContentView

BenGWeeks:
Oh - there's something in there! How do you get to fire off? Didn't think we had it implemented!

jb55:
it supports the old nostr:e:hexkey and nostr:p:hexkey thing but we should make it support npub and note

BenGWeeks:
OK cool. Happy to update if you like (we now have that part)? I like the method on the content view as well - looks like that would also resolve the damus_status issue I was having.

jb55:
Yeah I don't think nav is being done correctly, it uses a single stack which is probably wrong?

BenGWeeks:
Yeah, I believe so  ... but I'm no expert ... but it looks like one of those jobs you crack open and it ends up being quite a bit to sort it all out. I did actually start doing that with the design updates but soon realised we would have a merge night mare ;-)

Updated this to work where the previous code was. It's now going to the profile page for a string such as (my nPub):

`nostr://io.damus.nostr/npub1jutptdc2m8kgjmudtws095qk2tcale0eemvp4j2xnjnl4nh6669slrf04x`

However, it's not loading correctly, so must be missing something at the moment. Will continue looking into it. I think it's something to do with the fact I'm passing the nPub as a string, but the reference ID isn't that.

Have abstracted the URL handler into `DeepLinkManager` although I haven't removed `NostrLink` as could be

BenGWeeks:
@jb55, you mentioned you were working on the sharing links going out ... is that going via the website that redirects to `nostr:\\` if installed or directly to `nostr://`? What format are you using?

jb55:
I fixed profile and note loading in latest build! let me know what you think

BenGWeeks:
Oooh - lovely, let me go look.

BenGWeeks:
I see the share :-) In Safari that goes to [damus.io](https://damus.io) but then forwards to snort.social. Not sure why - but I think the we need to change the following to `damus.io` (and the corresponding value in `DeepLinkManager` class if you'#re still using that:

image

jb55:
that section is for uris. The feature we have now is universal links via domain association entitlement:

https://damus.io/.well-known/apple-site-association

If you click the link in the notes app, signal, telegram, etc it will open in damus.

Not sure why it doesn't work in some apps though.

jb55:
testing link here... https://damus.io/note1ws6wdg3rgmgue2sjxqp3gkmql70nsvm4fphsgl6nr7km2y57tzpsd0gy6w

BenGWeeks:
Do we want to /e/ and /p/ in the Url? (Saw that used but technically don't _need_).

jb55:
this isn't a problem anymore right?

Issue #257, opened 2023-01-05
EXC_BAD_ACCESS crashes development version


BenGWeeks:
Suspect this occurs as we're killing and re-opening the app frequently and no new contact events or something.

Will submit a PR for a resolution.

jb55:
old and likely not a problem anymore

PR #258, opened 2023-01-05
Fixes damus-io/damus#257

BenGWeeks:
ChangeLog-Changed: Added additional error handling to loading contact events (@npub1jutptdc2m8kgjmudtws095qk2tcale0eemvp4j2xnjnl4nh6669slrf04x)


jb55:
the if let should handle this?

BenGWeeks:
In my experience it fixed issue where app would crash if I opened/closed it too quickly yes. Could wrap a few more lines in that statement from memory but seemed to work fine as it was.

BenGWeeks:
(It would crash as events was nil, presumably because no new events from contacts as opened it too fast after previously closing).

PR #264, opened 2023-01-06
remove unused packages

aki-mizu:
Shimmer package was added to package dependencies in #56 but Shimmer.swift in Components was used instead.


jb55:
thanks!

aki-mizu:
Merged in 424fb5534301b65351522934b01feb8702bbcfb0

PR #267, opened 2023-01-06
Hide the NIP-05 Domain in Follow List

joelklabo:
Seemed to be always breaking the layout so I figured you can go to their profile if you want to check the actaul domain, check mark is probably enough in follow list. Might make sense for the Boosted view as well.

| Before      | After |
| ----------- | ----------- |
|  Simulator Screen Shot - iPhone 14 Pro - 2023-01-05 at 19 22 07    |  Simulator Screen Shot - iPhone 14 Pro - 2023-01-05 at 19 21 45  |

### DIFF
--- a/damus/Views/FollowingView.swift
+++ b/damus/Views/FollowingView.swift
@@ -24,7 +24,7 @@ struct FollowUserView: View {
             
                 VStack(alignment: .leading) {
                     let profile = damus_state.profiles.lookup(id: target.pubkey)
-                    ProfileName(pubkey: target.pubkey, profile: profile, damus: damus_state, show_friend_confirmed: false)
+                    ProfileName(pubkey: target.pubkey, profile: profile, damus: damus_state, show_friend_confirmed: false, show_nip5_domain: false)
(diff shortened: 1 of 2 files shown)


jb55:
Should we just use EventProfileName? not even sure why we have two different views for this.

joelklabo:
I think the EventProfileName requires some different parameters and also has the`@person` part there, which maybe would be good? I think the follow icon as well. But yeah you're right I'm not sure what should go where. Might need a designer to chime in on that one.. Not my strong suit

BenGWeeks:
I think we should remove the "little man" icon (generally) and add a "Follows you" to the profile screen.

BenGWeeks:
FYI I was working on a feature where you click the badge and the NIP-05 identifier slides out to be revealed to save space if that's something we're also interested in?

joelklabo:
I can resolve these conflicts if it seems like a good idea to merge this but if not I'll just let it go probably

jb55:
thanks! this is in the merge queue

PR #268, opened 2023-01-06
New feature: Image uploading/hosting

michaelhall923:
Hi all,

I've developed a feature for users to add images to their posts in-app. It works by sending the chosen image to an image host API (right now nostrimg.com is the only option) and appending the returned image URL to the end of the post.

It is flexible so that new APIs can easily be added simply by filling in some settings and implementing the uploadImage() function for the API. All of the API code can be found in Models/ImageHost.swift.

Also included in the PR is the ability for users to set a default image host in the config view.

I have also written an ImagePicker view which can be reused for other purposes in the app such as setting profile pic.

I had a hard time getting the layout to work well for the ReplyView, so for I've only enabled image uploading when creating top-level posts, but this can easily be changed by removing a couple lines of code.

Supports every image format I tested, including gifs. (It seems like the PHPickerViewController just converts still images to JPEG. It did that for webp, heif, png, etc.)

Other hosts worth looking into: nostr.build, imgur

Apologies if I did anything horribly wrong, I'm very new to iOS dev 😝
IMG_5610
IMG_5611

### DIFF
--- a/damus/Models/ImageHost.swift
+++ b/damus/Models/ImageHost.swift
@@ -1,0 +1,108 @@
+//
+//  ImageHost.swift
+//  damus
+//
+//  Created by Michael Hall on 01/04/23.
+//
+
+import Foundation
+import UIKit
+
+enum ImageHost: String, CaseIterable, Identifiable {
+    var id: String { self.rawValue }
+    
+    struct Model: Identifiable, Hashable {
+        var id: String { self.tag }
+        var tag: String
+        var displayName : String
+        var uploadLink : String
+        var clientId : String?
+    }
+    
+// TODO: implement nostr.build
+//    case nostr_build
+    case nostrimg
+    
+    var model: Model {
+        switch self {
+// TODO: implement nostr.build
+//        case .nostr_build:
+//            return .init(tag: "nostr_build", displayName: NSLocalizedString("nostr.build", comment: "Dropdown option label for image host Nostr.Build."),
+//                         uploadLink: "https://nostr.build/api/upload", clientId: "")
+        case .nostrimg:
+            return .init(tag: "nostrimg", displayName: NSLocalizedString("nostrimg.com", comment: "Dropdown option label for image host Nostrimg."),
+                         uploadLink: "https://nostrimg.com/api/upload", clientId: "401e1e2049ba44114a5f59e60315eb0fb0fd12beee1f08deb8bde773995b896e")
+        }
+    }
+    
+    static var allModels: [Model] {
+        return Self.allCases.map { $0.model }
+    }
+    
+    func uploadImage(image_data: Data, completion: @escaping (Result<URL, Error>) -> Void) {
+        switch self {
+// TODO: implement nostr.build
+//        case .nostr_build:
+//
+//            return
+            
+        case .nostrimg:
+            // Make API call to Nostrimg to upload image
+            let url = URL(string: self.model.uploadLink)
+
+            // generate boundary string using a unique per-app string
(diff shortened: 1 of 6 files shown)


BenGWeeks:
Nice :-) I like the idea of multiple options being available.

damus/Models/ImageHost.swift
```diff
@@ -0,0 +1,108 @@
...
+    
+    var model: Model {
+        switch self {
+// TODO: implement nostr.build
+//        case .nostr_build:
+//            return .init(tag: "nostr_build", displayName: NSLocalizedString("nostr.build", comment: "Dropdown option label for image host Nostr.Build."),
+//                         uploadLink: "https://nostr.build/api/upload", clientId: "")
+        case .nostrimg:
+            return .init(tag: "nostrimg", displayName: NSLocalizedString("nostrimg.com", comment: "Dropdown option label for image host Nostrimg."),
+                         uploadLink: "https://nostrimg.com/api/upload", clientId: "401e1e2049ba44114a5f59e60315eb0fb0fd12beee1f08deb8bde773995b896e")
```
jb55, line 34: I do not want to rely on any hardcoded integrations like this unless it's something that I know will keep working

jb55:
if I were to add something like this it would have to have a proper spec so that you can add/remove hosts without having anything hardcoded.

CH-SYR3:
Is this live? How do I update damus?

BenGWeeks:
Tricky as all hosts would have bespoke API. Bitcoin Jungle looked at an interesting solution using Torrent.

michaelhall923:
> if I were to add something like this it would have to have a proper spec so that you can add/remove hosts without having anything hardcoded.

As in, Damus calls APIs in one specific way, and it’s up to the hosts to adhere to that protocol?

michaelhall923:
> Tricky as all hosts would have bespoke API. Bitcoin Jungle looked at an interesting solution using Torrent.

That's a really cool idea. But it seems like it would make it hard to remove illegal content ?

leesalminen:
Wondering about standardizing on S3 compatible APIs? Users can use tons of services that conform to this standard now, and nostr specific services could expose a compatible endpoint.

BenGWeeks:
> Wondering about standardizing on S3 compatible APIs? 
> 
> 
> 
> Users can use tons of services that conform to this standard now, and nostr specific services could expose a compatible endpoint. 
> 
> 
> 
> Users would set their endpoint and keys in settings on the client. 
> 
> 
> 
> Clients could use signed URLs to upload directly. 

I though S3 was very Amazon specific?

But yes, users setting API settings sounds the trick. What was that torrent based solution you sent me?

leesalminen:
> > Wondering about standardizing on S3 compatible APIs?
> > Users can use tons of services that conform to this standard now, and nostr specific services could expose a compatible endpoint.
> > Users would set their endpoint and keys in settings on the client.
> > Clients could use signed URLs to upload directly.
> 
> I though S3 was very Amazon specific?
> 
> But yes, users setting API settings sounds the trick. What was that torrent based solution you sent me?

S3 has become such a behemoth that many (most?) other cloud object store providers offer a S3-compatible API so that a customer could migrate from AWS to whatever service with minimal hassle. See:

- https://www.backblaze.com/b2/docs/s3_compatible_api.html
- https://developers.cloudflare.com/r2/data-access/s3-api/api/
- https://docs.wasabi.com/docs/wasabi-api
- https://docs.oracle.com/en-us/iaas/Content/Object/Tasks/s3compatibleapi.htm

PR #269, opened 2023-01-06
Update readme markdown format guides in Formatting Notes

aki-mizu:
In iOS for strikethrough only one tilde symbol is needed before and after words


jb55:
thx

PR #271, opened 2023-01-06
Fix TabView loading view on scroll

OlegAba:
Before             |  After
 |

### DIFF
--- a/damus/ContentView.swift
+++ b/damus/ContentView.swift
@@ -93,9 +93,9 @@ struct ContentView: View {
     var PostingTimelineView: some View {
         VStack {
             TabView(selection: $filter_state) {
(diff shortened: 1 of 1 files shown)


jb55:
Thanks! 5f22a7691ff7202f398fd57d0b4d0c52900ba844

jb55:
I'm not sure what this fixes and it only makes switching tabs more buggy so I had to revert it. the before/after in your comment is not loading for me.

OlegAba:
Try horizontally scrolling (not use the TabBar) after you revert the changes. I see the tabBar bug on initial load - I'll fix it and resubmit. Can you see the before/after below?

Before             |  After
 |

PR #273, opened 2023-01-06
Implement sidebar

BenGWeeks:
Hi @jb55,

Are you interested in a sidebar we can add further options with time?

image

image

TODO:

- Implement fading background
- Hide the sidebar when navigation tabs clicked at the bottom
- Move relays to their separate view controller

ChangeLog-Changed: Implement new sidebar (@npub1jutptdc2m8kgjmudtws095qk2tcale0eemvp4j2xnjnl4nh6669slrf04x)


BenGWeeks:
Ok, I'll carry on with it then :-)

yeg0rpetrov:
This is good one!

BenGWeeks:
> This is good one!

Thought you might like this one ;-)

jb55:
probably needs a slide animation of some kind...

BenGWeeks:
Has but set to 0.2 seconds so maybe a bit fast?

jb55:
Yeah maybe, it doesn't look like it animates at all for me. Also not sure about the purple text.

BenGWeeks:
Maybe it's the `NavigationView` affecting it. The background should have an opacity too but can't see through it for some reason. If you put the `SideMenuView` outside of the `NavigationView` then the `NavigationLink`s don't work ...

BenGWeeks:
OK, I think this is ready for review @jb55 :-)

image

BenGWeeks:
image

BenGWeeks:
image

BenGWeeks:
FYI I'd love to see the Damus logo at the centre middle on the `Home` screen and a filter option on the top-right where the settings icon previously was (my understanding is the toolbar at the top of the screen is often related to functionality for the current page you are on, e.g. filter for home, settings for DMs, etc.).

BenGWeeks:
In the future you could potentially add more features, such as:

image

FYI There seems to be a maximum of 10 `NavigationLinks` possible within a `NavigationView` (hence no "Relays" in the screenshot above)!?

jb55:
thanks! I merged this. there is only one weird artifact that I see which is the damus icon disappearing when the sidebar is opened.

BenGWeeks:
Ah yes, there is a challenge with `NavigationView` and/or `ToolBars` and zIndex in that they always appear over the top. Maybe add some slide affect so it looks like it's supposed to the that or specifically only hide the title and profile image?

PR #275, opened 2023-01-07
Make NIP-05 Domain Tappable

joelklabo:
This makes tapping the domain shown after the verified badge take you to that site

Simulator Screen Recording - iPhone 14 Pro - 2023-01-06 at 16 07 42


jb55:
neat

PR #276, opened 2023-01-07
Update README.md

paulrollo:
The backticks around the code markdown example need escaped with two backticks and a space.


jb55:
thanks!

PR #277, opened 2023-01-07
Update Event Bar Icons and Labels to Stay Put When Interacting

joelklabo:
I noticed the event view changes size a little bit when you "like" a note. I tried using the same emoji but in grayscale when not liked so the sizes match. Just an idea but it's less jumpy which is nice I think.

Update:
Updated Shaka and Boost to match

| Before      | After |
| ----------- | ----------- |
|    Simulator Screen Recording - iPhone 14 Pro - 2023-01-06 at 19 39 56  | Simulator Screen Recording - iPhone 14 Pro - 2023-01-10 at 19 36 07 |


BenGWeeks:
Personally, I think the highlighted version should be the coloured version of the left (unfilled) to be consistent.

joelklabo:
Yeah, makes sense. Where did those assets come from? Or do you know who made them? Maybe we could ask, I don't think they filled in version is in the project at the moment.

BenGWeeks:
See #226

joelklabo:
Nice, I think we just need one version that's a PDF, which will fix the scaling issue too.

joelklabo:
Hey @elidyweaver would you be able to share the PDF or SVG version of the outline Shaka and the filled Shaka? I think that's all we need.

BenGWeeks:
I see GitHub mobile app has tabs at bottom that fill and color which is quite cool actually.

joelklabo:
Yeah, Figma definitely does. I've used it for that before. If you make it let me know if I can help with exporting to PDF.

BenGWeeks:
Here it is. However, I have a better understanding of where @jb55 wants to take it so designs need to be updated (Sasha was doing this) e.g. with the shaka hand.

Password: keypair
(If you log in, you can leave the most welcome comments)

BenGWeeks:
But some other posts on this (can't find to hand) and a separate Figma file with JUST the icon sizes and variations in that too.

BenGWeeks:
See #226. FYI @elidyweaver @joelklabo

elidyweaver:
> Hey @elidyweaver would you be able to share the PDF or SVG version of the outline Shaka and the filled Shaka? I think that's all we need.

The pdf or svg & color adjustments should be no problem. 
This was the original file: 
https://www.figma.com/file/f2xhDQVpfeV46UA0Hrpt7e/design-explorations?node-id=3%3A2&t=khcg5VtpIatStfxV-1

I will update it to be able to compare and decide on options.
The update will include:
Damus colors
Gradient option
Updated size (@BenGWeeks was working on a 24x24 pixels and I had a 20x20 - that might be the issue)

joelklabo:
I think we need a version of the filled Shaka that isn't an emoji because that's `text` basically and it won't match the outline version. Does that make sense? @elidyweaver @BenGWeeks

BenGWeeks:
Agree, should be consistent (although no particular preference as to how it is highlighted although I am a fan of the gradient).

joelklabo:
Yeah, we could start with just the outline and filled versions (same size) and then update it from there if we want. But I don't have a strong preference either. The filled one can be all white or what're too btw b/c we can tint it in code

joelklabo:
Ok I found your full and outline assets @elidyweaver and exported them as PDFs, see the result here:
Simulator Screen Recording - iPhone 14 Pro - 2023-01-07 at 12 48 09

BenGWeeks:
Nice :-) What's the plan, different colour per action icon?

joelklabo:
Not sure I know Boost is green so I don't think they are all supposed to be the same color? Happy to change though

BenGWeeks:
Yeah, i was just checking what Twitter did :-)

elidyweaver:
Glad you could use those, @joelklabo 
The more colors you have, the busier it can get. 

Here's an example of color options. I don't yet have a light version. Do any of you have a sample screenshot of the current light mode version I could use?
(I squeezed the lightning into the screenshot.)
colors

Also, could the shaka be an animated gif?

elidyweaver:
In terms of the animation, perhaps something like this: 
shaka

joelklabo:
@elidyweaver yeah that's cool! Definitely doable. Might need to get more agreement though so we could add in a subsequent small PR.

BenGWeeks:
> Perhaps something like this for animation. 
> 
> It's on an infinite loop so you can see it. It looks pixelated, will try to smooth out. 
> 
> 
> 
> shaka-5
> 
> 

Lightning will be an option just like the others FYI.

joelklabo:
Yeah, might make sense to update these all at once. Whatever lightning icon is used will have the same things to consider as the Shaka

BenGWeeks:
There's also iconography used in some of the other PRs.

I can see the challenge in open source development of being hesitant of going down the custom iconography route due to the issue of slowing down new features if custom icons not available.

jb55:
testing this again

jb55:
when you click the shaka it seems to move to the right

jb55:
I just realized it currently does this as well :P

joelklabo:
Yeah, when the label gets added it moves over, I was thinking we could have a hidden label there and then show it so it wouldn't move, or wouldn't move as much

joelklabo:
Should I close this down you think? Right now it just updating the Shaka to PDF icons. I thinks it's an improvement but maybe the design is going a different direction.

joelklabo:
Actually I think I figured this out. We can use `.overlay` to put the text above the Shaka so it doesn't move:
Simulator Screen Recording - iPhone 14 Pro - 2023-01-10 at 18 49 09

I'll update the description too

joelklabo:
I updated the Boost button to match and moved the labels to the right side to match Twitter.

Simulator Screen Recording - iPhone 14 Pro - 2023-01-10 at 19 36 07

joelklabo:
Not sure if this is worth rebasing anymore?

BenGWeeks:
I'm waiting for direction on iconography before trying to change that.

There is certainly benefits in keeping to SF Symbols for example as there are so many and makes development quicker.

Having said that, there are certain iconography (like "Shaka" hand that isn't there) and if we can have a full library of imagery that we need or likely WILL need that would be awesome to give a more branded look.

FYI @elidyweaver @SashaPavlova.

jb55:
> Not sure if this is worth rebasing anymore?

I'll take another look at this

joelklabo:
It's mainly this: 
image

It's probably faster to just redo it on master it wasn't a huge change

jb55:
I merged it fine and it looks great

jb55:
should we make it purple?

joelklabo:
Yeah,the label too I guess?

jb55:
I'll do it!

PR #284, opened 2023-01-08
Fix localization bugs, add localization string comments, and internationalize sats formatting

tyiu:


tyiu:
I amended the 3rd / last commit to handle the singular case where it's formatting `1 sat` (previously it was `1 sats`)

damus/Models/Mentions.swift
```diff
@@ -187,12 +187,22 @@ enum Amount: Equatable {
...
+            let numberFormatter = NumberFormatter()
+            numberFormatter.numberStyle = .decimal
+            numberFormatter.minimumFractionDigits = 0
+            numberFormatter.maximumFractionDigits = 3
+            numberFormatter.roundingMode = .down
+
+            let sats = NSNumber(value: (Double(amt) / 1000.0))
+            let formattedSats = numberFormatter.string(from: sats) ?? sats.stringValue
+
+            if formattedSats == numberFormatter.string(from: 1) {
```
tyiu, line 201: I amended the 3rd / last commit to handle the singular case where it's formatting `1 sat` (previously it was `1 sats`)

damus/Models/Mentions.swift
```diff
@@ -187,12 +187,22 @@ enum Amount: Equatable {
...
+                return NSLocalizedString("\(formattedSats) sat", comment: "Amount of 1 sat.")
             }
-            return "\(amt / 1000) sats"
+
+            return NSLocalizedString("\(formattedSats) sats", comment: "Amount of sats.")
```
tyiu, line 205: Amounts like `1234567` msats are now formatted as `1,234.567 sats` in US English and something like `1.234,567 sats` in German in Germany, for example.

Previously, there were no delimiters or fractions, so you just got `1234 sats`.

damus/Views/EditMetadataView.swift
```diff
@@ -160,9 +160,9 @@ struct EditMetadataView: View {
                     Text("NIP-05 Verification")
                 }, footer: {
                     if let parts = nip05_parts {
-                        Text(String.localizedStringWithFormat("'%@' at '%@' will be used for verification", parts.username, parts.host))
```
tyiu, line 163: Using the strings in this way does not make it localizable, so in my machine-translated Spanish tests, they were appearing as English.

tyiu:
Ended up stacking a few more commits to basically get master in a ready state for Transifex.

damus.xcodeproj/project.pbxproj
```diff
@@ -738,11 +739,12 @@
 			};
 			buildConfigurationList = 4CE6DEDE27F7A08100C66700 /* Build configuration list for PBXProject "damus" */;
 			compatibilityVersion = "Xcode 13.0";
-			developmentRegion = en;
```
tyiu, line 741: Using `en` as the development locale makes it ambiguous. If we ever introduce strings like `color`/`colour` or `favorite`/`favourite`, it'll be inconsistent. Since Apple intentionally uses `en-US`, we'll use `en-US` for Damus as well.

damus.xcodeproj/project.pbxproj
```diff
@@ -738,11 +739,12 @@
...
+			developmentRegion = "en-US";
 			hasScannedForEncodings = 0;
 			knownRegions = (
-				en,
 				Base,
+				"es-419",
```
tyiu, line 746: `es-419` is Spanish (Latin America). That should hopefully cover a large enough market. If there are differences in translations for specific versions of Spanish within Latin American regions, we can create target locales for those too.

damus.xcodeproj/project.pbxproj
```diff
@@ -1027,6 +1031,7 @@
 				ONLY_ACTIVE_ARCH = YES;
 				SDKROOT = iphoneos;
 				SWIFT_ACTIVE_COMPILATION_CONDITIONS = DEBUG;
+				SWIFT_EMIT_LOC_STRINGS = YES;
```
tyiu, line 1034: Xcode 13 seems to turn this on by default for new projects, but Damus didn't have it, so I turned it on.
https://developer.apple.com/documentation/xcode-release-notes/xcode-13-release-notes

> The new Use Compiler to Extract Swift Strings build setting invokes the Swift compiler to accurately extract string interpolations and string literals from Text(), String(localized:), AttributedString(localized:) initializers, SwiftUI’s LocalizedStringKey, and Foundation’s StringLocalizationKey. Enabling this setting compiles all localized targets in your project when exporting and importing for localization. This build setting is enabled by default on all new projects, but can be opted-in for existing Swift projects. (58023398)

damus.xcodeproj/project.pbxproj
```diff
@@ -1035,6 +1040,7 @@
 			isa = XCBuildConfiguration;
 			buildSettings = {
 				ALWAYS_SEARCH_USER_PATHS = NO;
+				CLANG_ANALYZER_LOCALIZABILITY_NONLOCALIZED = YES;
```
joelklabo, line 1043: Does this show warnings for non-localized strings?

joelklabo (commented):
LGTM! That's a lot to do.  I'm curious if it's possible to put the comments in the `.strings` files? In another app I've worked on they are there and we just pass `nil` for the comment when calling `NSLocalizedString` but it removes a lot of text from the code which is nice. 

I'm not sure if that works with this setup, it may have been something custom.

Pretty cool!

transifex.yml
```diff
@@ -0,0 +1,13 @@
...
+    file_format: XLIFF
+    source_language: en_US
+    source_file: damus Localizations/en-US.xcloc/Localized Contents/en-US.xliff
+    translation_files_expression: damus Localizations/<lang>.xcloc/Localized Contents/<lang>.xliff
+
+  settings:
+    pr_branch_name: transifex/<br_unique_id>
+    language_mapping:
+      en_US: en-US
+      es_419: es-419
```
tyiu, line 13: Amended a fix on the last commit. Turns out Transifex uses locale codes in the form of `es_419`, for example, but Xcode uses `es-419`. i.e. Transifex uses `_` (underscore) to separate language and region/variant, and Xcode uses `-` (dash). We can use this mapping to convert between the two. It's a pain to keep this list up to date, but not sure if there are alternative options.

damus.xcodeproj/project.pbxproj
tyiu: I haven't figured out how to trigger it. I tried removing one of the NSLocalizedString macros but I didn't see any warnings.

joelklabo: Yeah, I've had the same experience. Can't hurt though if it starts working eventually.

tyiu:
> LGTM! That's a lot to do. I'm curious if it's possible to put the comments in the `.strings` files? In another app I've worked on they are there and we just pass `nil` for the comment when calling `NSLocalizedString` but it removes a lot of text from the code which is nice.
> 
> I'm not sure if that works with this setup, it may have been something custom.
> 
> Pretty cool!

@joelklabo I'm typically a backend developer and only occasionally dabble in iOS. As far as I can tell, there's three ways we can do this:

1. How it's set up in this PR.
Pro: Easy to use for the developer making the change
Cons: (a) More verbosity in code. (b) Duplicate strings in different parts of code could conflict if they have different meanings / translations depending on the context.

2. Whoever is introducing new or changing existing strings still adds the `NSLocalizedString` macro with the source English string in code with no comment, but they need to run `genstrings` like what this [StackOverflow](https://stackoverflow.com/a/62093002) answer suggests, and then supply the comments in the `.strings` files.
Pro: Less verbosity in code
Cons: (a) Developer needs to remember to run `genstrings` to add comments. (b) Duplicate strings in different parts of code could conflict if they have different meanings / translations depending on the context.

3. Use the `NSLocalizedString` macro but reference just a key instead of actually using the source English string in code.
Pro: (a) Less verbosity in code. (b) No duplicate strings unless there's a key collision for some reason.
Con: (a) Developer needs to run `genstrings` to add the source English string and comments. (b) Less clear about what string is being used when reading code until you look it up in the .strings file.

I think I'm favoring sticking with the current approach in option 1, but can be convinced otherwise. I'm amenable to other options if we can make it easier on the developer somehow.

damus/en-US.lproj/Localizable.stringsdict
```diff
@@ -0,0 +1,60 @@
...
+		<key>NSStringLocalizedFormatKey</key>
+		<string>··· %#@NOTES@ ···</string>
+		<key>NOTES</key>
+		<dict>
+			<key>NSStringFormatSpecTypeKey</key>
+			<string>NSStringPluralRuleType</string>
+			<key>NSStringFormatValueTypeKey</key>
+			<string>d</string>
+			<key>zero</key>
+			<string>0 other notes</string>
```
tyiu, line 52: I introduced a bug where these strings omitted the word `other`. I amended the commits to re-add them, hence the force push.

joelklabo:
> > LGTM! That's a lot to do. I'm curious if it's possible to put the comments in the `.strings` files? In another app I've worked on they are there and we just pass `nil` for the comment when calling `NSLocalizedString` but it removes a lot of text from the code which is nice.
> > I'm not sure if that works with this setup, it may have been something custom.
> > Pretty cool!
> 
> @joelklabo I'm typically a backend developer and only occasionally dabble in iOS. As far as I can tell, there's three ways we can do this:
> 
> 1. How it's set up in this PR.
>    Pro: Easy to use for the developer making the change
>    Cons: (a) More verbosity in code. (b) Duplicate strings in different parts of code could conflict if they have different meanings / translations depending on the context.
> 2. Whoever is introducing new or changing existing strings still adds the `NSLocalizedString` macro with the source English string in code with no comment, but they need to run `genstrings` like what this [StackOverflow](https://stackoverflow.com/a/62093002) answer suggests, and then supply the comments in the `.strings` files.
>    Pro: Less verbosity in code
>    Cons: (a) Developer needs to remember to run `genstrings` to add comments. (b) Duplicate strings in different parts of code could conflict if they have different meanings / translations depending on the context.
> 3. Use the `NSLocalizedString` macro but reference just a key instead of actually using the source English string in code.
>    Pro: (a) Less verbosity in code. (b) No duplicate strings unless there's a key collision for some reason.
>    Con: (a) Developer needs to run `genstrings` to add the source English string and comments. (b) Less clear about what string is being used when reading code until you look it up in the .strings file.
> 
> I think I'm favoring sticking with the current approach in option 1, but can be convinced otherwise. I'm amenable to other options if we can make it easier on the developer somehow. It's also possible I've missed other alternative options.

Yeah, that makes sense to me. Just different trade-offs but I don't have a strong opinion that it should be changed from what you have. There actually aren't a ton of strings in the app so it's probably not that bad just to have it inline like you do.

jb55:
thanks!

PR #286, opened 2023-01-08
Make DMs only take up 80% of screen width to match iMessages

realprogrammersusevim:
DMs now take up only 80% of the available screen width so it looks much more like iMessages. Simply added spacers at each end of the message depending on who's message it was.
Screenshot 2023-01-07 at 7 54 04 PM


jb55:
thanks!

PR #289, opened 2023-01-08
Relay filtering

BenGWeeks:
Hi @jb55,

Let me know if you're interested in relay filtering on the homepage and I'll tidy this one up so it's ready:

image

TODO:

- Update homepage results to only the selected relays.
- Move posts/replies into this filter area too.

Either in this PR or another one, we could add the ability to choose the relay you wish to post messages to.


BenGWeeks:
Yeah, for me this is more about setting expectations and behaviour that the future intention is not all relays will have ALL messages and for a way to get rid of the spam (that will come to free relays).

We could also put other filtering in here too.

jb55:
going to experiment with this, I would like an easy way to connect/disconnect from relays. maybe this would be a good way to do that.

BenGWeeks:
We actually have this on the search now I notice.

I think it would be good to have this on "Home" too.

However, I think the concept could be improved somewhat - just not sure how!? For instance, I would like to be able to easily switch from seeing everything to just seeing a relay that maybe I set up for my company or a family relay. FYI @alltheseas @robagreda @SashaPavlova

alltheseas:
I think there could be different configurable relay views (or relay sets). Refer to https://github.com/damus-io/damus/issues/453.

Maybe most zapped notes view is more important than relay sets.

BenGWeeks:
> I think there could be different configurable relay views (or relay sets). Refer to #453.
> 
> Maybe most zapped notes view is more important than relay sets.

Yeah, heard relay sets being discussed. Remain to be convinced this is the most understandable way?

The "most zapped" view I think relates to a separate idea where you can select your own algo. These algos could be customized and/or shared eventually.

alltheseas:
Yes, there is not enough definition or consensus on relay views, and filtering.

There is consensus that most zapped notes, most zapped comments is a view that we want to support.

My recommendation is to implement the most zapped views before taking on relay filtering outside the universe.

PR #291, opened 2023-01-09
Beautify home navigation bar title

BenGWeeks:
Hi @jb55,

A very simple tweak with this one, to display a small logo on the homepage (we have 1x, 2x and 3x versions now). I noticed that Twitter did this and thought it looked quite nice.

image

Also updated from `.navigationBarTitle` to preferred `.toolbar`.

This should go quite well with the relay filter (new filter icon at the top right) and the left sidebar.

If we move the Posts / Posts & Replies to the filter pop-up pane, that would also give more space at the top.

Cheers,

Ben


jb55:
oh nice

damus/ContentView.swift
```diff
@@ -168,7 +168,20 @@ struct ContentView: View {
...
+        .navigationBarTitleDisplayMode(.inline)
+        .toolbar {
+            ToolbarItem(placement: .principal) {
+                if selected_timeline == .home {
+                    Image("damus-home")
+                    .resizable()
+                    .frame(width:30,height:30)
+                    .shadow(color: Color("DamusPurple"), radius: 2)
+                } else {
+                    Text("Global")
```
tyiu, line 180: @BenGWeeks @jb55 

I think this line introduced a bug where it shows `Global` instead of `DM` in the toolbar for the DM view.

I'm working on a PR to fix this now.

tyiu: https://github.com/damus-io/damus/pull/321

PR #292, opened 2023-01-09
Fix Markdown Parsing in Profile

joelklabo:
It looks like we added some code to add markdown for email addresses (`mailto:`) and links without `https://` but that also make explicit markdown links not work because it tried to add markdown inside the parenthesis of the markdown link.

This allows you to use markdown links in the profile and would require you to use markdown if you want to something like `damus.io` to appear as a link. But, I think that's probably expected. I may be misunderstanding the purpose here but everything seems to work for me except the links like `damus.io` but I think people can just use markdown for that?

FYI @lionello

### DIFF
--- a/damusTests/MarkdownTests.swift
+++ b/damusTests/MarkdownTests.swift
@@ -21,44 +21,32 @@ class MarkdownTests: XCTestCase {
     }
 
     func test_convert_link() throws {
-        let helper = Markdown()
-        let md = helper.process("prologue https://nostr.build epilogue")
+        let md = Markdown.parse(content: "prologue https://nostr.build epilogue")
         let expected = try AttributedString(markdown: "prologue [https://nostr.build](https://nostr.build) epilogue", options: md_opts)
         XCTAssertEqual(md, expected)
     }
-
-    func test_convert_link_no_scheme() throws {
-        let helper = Markdown()
-        let md = helper.process("prologue damus.io epilogue")
-        let expected = try AttributedString(markdown: "prologue [damus.io](https://damus.io) epilogue", options: md_opts)
+    
+    func test_no_convert_markdown_link() throws {
+        let md = Markdown.parse(content: "prologue [link](https://nostr.build) epilogue")
+        let expected = try AttributedString(markdown: "prologue [link](https://nostr.build) epilogue", options: md_opts)
         XCTAssertEqual(md, expected)
     }
-
-    func test_convert_links() throws {
-        let helper = Markdown()
-        let md = helper.process("prologue damus.io https://nostr.build epilogue")
-        let expected = try AttributedString(markdown: "prologue [damus.io](https://damus.io) [https://nostr.build](https://nostr.build) epilogue", options: md_opts)
+    
+    func test_no_convert_with_emoji() throws {
+        let md = Markdown.parse(content: "test link w/ emoji ❤️ in a [string](https://nostr.build)")
+        let expected = try AttributedString(markdown: "test link w/ emoji ❤️ in a [string](https://nostr.build)", options: md_opts)
         XCTAssertEqual(md, expected)
     }
 
     func test_convert_http() throws {
-        let helper = Markdown()
-        let md = helper.process("prologue http://example.com epilogue")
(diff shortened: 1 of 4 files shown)


lionello:
This is really a spec question: is the profile always markdown, or plaintext? If it's always markdown, then this change seems correct to me. But if it's supposed to be plaintext, then I suppose we need another fix to detect markdown'ed links before we use the link detector. That doesn't seem like it's hard to do either: when we find a link we check whether there's parantheses around it. If so, don't add markdown.

joelklabo:
> This is really a spec question: is the profile always markdown, or plaintext? If it's always markdown, then this change seems correct to me. But if it's supposed to be plaintext, then I suppose we need another fix to detect markdown'ed links before we use the link detector. That doesn't seem like it's hard to do either: when we find a link we check whether there's parantheses around it. If so, don't add markdown.

Yeah, I think being able to use markdown in you profile is pretty nice. But maybe it should be noted in the UI, like add a line that says `markdown supported` or something.

I actually went down the road of trying to detect links that were already markdown formatted like you mentioned. Just looking for paranthesis on each side. But, ran into some issues with emoji etc making it difficult to not get an exception for out of bounds. But it's definitely a possible way to go. Strings in Swift are just kind of tricky so I though I might end up adding an unexpected crash or something.

damusTests/MarkdownTests.swift
```diff
@@ -21,47 +21,35 @@ class MarkdownTests: XCTestCase {
...
         XCTAssertEqual(md, expected)
     }
-
-    func test_convert_links() throws {
-        let helper = Markdown()
-        let md = helper.process("prologue damus.io https://nostr.build epilogue")
-        let expected = try AttributedString(markdown: "prologue [damus.io](https://damus.io) [https://nostr.build](https://nostr.build) epilogue", options: md_opts)
+    
+    func test_no_convert_with_emoji() throws {
+        let md = Markdown.parse(content: "test link w/ emoji ❤️ in a [string](https://nostr.build)")
```
lionello, line 36: Did you mean to put the heart within the `[…]`?

joelklabo: Yeah, emojis can mess up the character counts so I think it's important to test. They don't always fit into one utf16 character so the counts can be off.

lionello: got it! We should make sure we have a character with unicode surrogates! Got this one from Wikipedia: [𐐷](https://en.wikipedia.org/wiki/%F0%90%90%8F)

joelklabo: yeah, good call

jb55:
I would prefer if markdown links were not possible but it seems like we're kind of stuck with it? many things like mentions, hashtags, etc use markdown links internally. maybe there's another way to linkify things with attributed strings that I don't know about.

The markdown thing is not spec'd and it looks like crap on many clients. If there was a proper markdown NIP that put plaintext in content and markdown in a tag then that would be much better.

joelklabo:
Yeah, you can definitely insert links with just attributed strings. Might make sense to just remove markdown support entirely then? And handle inserting links with attributed strings. I don't know if many people use the other features of markdown like formatting etc. But definitely seems more simple to have it everywhere or nowhere to me.

lionello:
+1 for "everywhere or nowhere". Markdown in notes definitely feels excessive. Markdown can also hide the actual domain that a link points to, ie. could do something like `[google.com](https://blah.com)` which might open up a can of worms.

jb55:
> Yeah, you can definitely insert links with just attributed strings. Might make sense to just remove markdown support entirely then? And handle inserting links with attributed strings. I don't know if many people use the other features of markdown like formatting etc. But definitely seems more simple to have it everywhere or nowhere to me.

yes if someone knows how to do this I would appreciate it

joelklabo:
So you would want to remove markdown parsing altogether and do links, mentions, note-mentions etc. just with attributed string?

joelklabo:
Here's a PR for moving to `AttributedString` instead: #301

joelklabo:
Closing this in favor of #301

PR #293, opened 2023-01-09
Hide Recommended Relays Section if Empty

joelklabo:
If you've already added all the recommended relays, hide that section when it's empty:

| Before      | After |
| ----------- | ----------- |
|  Simulator Screen Shot - iPhone 14 Pro - 2023-01-08 at 21 47 41 |  Simulator Screen Shot - iPhone 14 Pro - 2023-01-08 at 21 47 51 |


jb55:
thx!

PR #296, opened 2023-01-10
Fix for relay lists being incorrect when adding a relay

constcast-glitch:


constcast-glitch:
When adding a relay an old event copy was used. This change makes sure the up to date contacts event is used. Fixes issues with the relay list behaving odd when adding and removing relays.

jb55:
oh interesting, I would except swift to do the right thing here... weird

PR #297, opened 2023-01-10
feat(ci): Publish PR notifications to Nostr

bndw:
Sends pull request notifications to Nostr using Github Workflows and noscl.

To setup, you'll need to add a `NOSTR_KEY` private key secret to this repository's Settings:
```
Settings tab > Secrets and Variables
```

- Test PR: https://github.com/bndw/nostrpr/pull/5
- Resulting post to Nostr:
image

### DIFF
--- a/.github/workflows/pr-to-nostr.yaml
+++ b/.github/workflows/pr-to-nostr.yaml
@@ -1,0 +1,29 @@
+name: Nostr PR Notifier
+on: 
+  pull_request:
+    types: [opened, reopened]
+jobs:
+  Notify:
+    runs-on: ubuntu-latest
+    env:
+      # Set in Github repo > Settings > Secrets and Variables
+      NOSTR_KEY: ${{ secrets.NOSTR_KEY }}
+      # Event vars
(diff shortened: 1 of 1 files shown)


jb55:
haha perfect thx

.github/workflows/pr-to-nostr.yaml
```diff
@@ -0,0 +1,29 @@
...
+      PR_BODY: ${{ github.event.pull_request.body }}
+    steps:
+      - name: Setup Go
+        uses: actions/setup-go@v2
+        with:
+          go-version: '1.19.0'
+      - run: go install github.com/fiatjaf/noscl@latest
+      - run: which noscl
+      - run: noscl setprivate $NOSTR_KEY
+      - run: noscl relay add wss://relay.damus.io
```
bndw, line 24: Probably want to move these into a variable, but then you'll need to split on a comma/etc. Figured I'd land this first and we can go from there.

joelklabo:
I wonder if we could add a map of GitHub username to nostr npub so you would get mentions on them.

PR #301, opened 2023-01-10
Use Attributed Strings for Linking Intead of Markdown

joelklabo:
Using `AttributedString` to handle links, user mentions, note mentions, and hashtags. Colors are just for demonstration. Tapping the links does route you to the correct thing

Still need to clean this up a little but wanted to get feedback before spending time on that.

 | Colors just for showing different link types (they are all purple in branch) |
| ----------- |
| image
  |


joelklabo:
Looks like we need to set the private key for the notifier btw @jb55

jb55:
awesome

jb55:
I see we have 2 link parsers now? I handle all the parsing tasks in C...

I also noticed links are broken with these changes...

joelklabo:
> I see we have 2 link parsers now? I handle all the parsing tasks in C...
> 
> 
> 
> I also noticed links are broken with these changes...

Yeah, there's probably a better way. I think when you parse into Blocks in C maybe that should only be for things that don't show as text. Invoices, images, etc. and the text parsing could be parsed directly to 'AttributedString' so the content would always be an 'AttributedString'. 

But that would mean two parsers or separate parsers I guess. 

I tested the links I thought, sorry about that.

joelklabo:
Going to close this in favor of this new pr https://github.com/damus-io/damus/pull/398

PR #302, opened 2023-01-10
Made banner image customizable by the user

jj0b:
Added new `banner` metadata field to hold a user settable banner image url. 

Updated the [profile edit screen](https://github.com/jj0b/damus/blob/master/damus/Views/EditMetadataView.swift) to have the same top section with the banner image and profile image (picture in the metadata) as is shown on the profile screen and added a new form field to allow updating the `banner` value.

According to the [NIP-28, Kind 41 docs](https://github.com/nostr-protocol/nips/blob/3423a6dfbc36c2039bb32a32e34186af875f820a/28.md#kind-41-set-channel-metadata) adding additional metadata fields _is_ allowed, so I think adding this `banner` field should be ok. I assume other clients will just ignore it unless they want to adopt it too.

### DIFF
--- a/damus/Models/HomeModel.swift
+++ b/damus/Models/HomeModel.swift
@@ -579,6 +579,15 @@ func process_metadata_event(profiles: Profiles, ev: NostrEvent) {
     }
     
     notify(.profile_updated, ProfileUpdate(pubkey: ev.pubkey, profile: profile))
+    
+    let banner = tprof.profile.banner ?? ""
+    if let _ = URL(string: banner) {
+        DispatchQueue.main.async {
+            notify(.profile_updated, ProfileUpdate(pubkey: ev.pubkey, profile: profile))
+        }
+    }
+    
+    notify(.profile_updated, ProfileUpdate(pubkey: ev.pubkey, profile: profile))
 }
 
 func robohash(_ pk: String) -> String {
--- a/damus/Models/HomeModel.swift
+++ b/damus/Models/HomeModel.swift
@@ -578,8 +578,6 @@ func process_metadata_event(profiles: Profiles, ev: NostrEvent) {
         }
     }
     
-    notify(.profile_updated, ProfileUpdate(pubkey: ev.pubkey, profile: profile))
-    
     let banner = tprof.profile.banner ?? ""
     if let _ = URL(string: banner) {
         DispatchQueue.main.async {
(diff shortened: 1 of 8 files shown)


jb55:
whoa! amazing

damus/Views/BannerImageView.swift
```diff
@@ -0,0 +1,74 @@
...
+
+struct InnerBannerImageView: View {
+    let url: URL?
+    let pubkey: String
+
+    var body: some View {
+        ZStack {
+            Color(uiColor: .systemBackground)
+            
+            AsyncImage(url: url) { image in
```
jb55, line 19: We want to use KingFisher here, AsyncImages are not cached

damus/Models/HomeModel.swift
```diff
@@ -579,6 +579,15 @@ func process_metadata_event(profiles: Profiles, ev: NostrEvent) {
     }
     
     notify(.profile_updated, ProfileUpdate(pubkey: ev.pubkey, profile: profile))
```
jb55, line 581: this line needs to be removed, we don't need to update profiles twice

jb55 (changes requested):

damus/Views/BannerImageView.swift
jj0b: Yeah, I wondered about that. I initially tried to use KingFisher but couldn't find a way to make a KFAnimatedImage or KFImage scale correctly like setting the resizable property on a UIImage (or AsyncImage) but I'll take another stab at it. Thanks for the feedback!

damus/Models/HomeModel.swift
jj0b: Fixed. :)

damus/Views/BannerImageView.swift
jj0b: Got it! Updated to use KFAnimatedImage for the banner image.

jj0b:
@jb55 requested changes implemented. Thanks again for the feedback and for all your work on Damus!

jb55:
just noticed a bug... for some reason my banner image shows up on other peoples profile..

jj0b:
> just noticed a bug... for some reason my banner image shows up on other peoples profile..

Hmm. I'll see if I can repro and fix. Thanks!

PR #307, opened 2023-01-11
Fix spacing on transifex.yml

tyiu:
Transifex wouldn't allow installation of integration otherwise.


jb55:
sorry for the delay, this is in my merge queue now

PR #309, opened 2023-01-11
Fix Drawer Layout Issues Using ZStack

joelklabo:
Fixed some layout issues with the drawer and don't need to hide the navigation anymore by using `ZStack` instead of `.overlay`. FYI @BenGWeeks 

| Before      | After |
| ----------- | ----------- |
| Simulator Screen Recording - iPhone 14 Pro - 2023-01-11 at 11 37 02  |  Simulator Screen Recording - iPhone 14 Pro - 2023-01-11 at 11 32 31 |

### DIFF
--- a/damus/ContentView.swift
+++ b/damus/ContentView.swift
@@ -245,16 +245,16 @@ struct ContentView: View {
                                                         .font(.callout)
                                                         .foregroundColor(.gray)
                                                 }
-
+                                                
                                             }
                                         }
                                     }
-
+                                
                             }
                         }
                     }
                     .navigationViewStyle(.stack)
-                
+                    
                     TabBar(new_events: $home.new_events, selected: $selected_timeline, isSidebarVisible: $isSideBarOpened, action: switch_timeline)
                         .padding([.bottom], 8)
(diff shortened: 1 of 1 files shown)


joelklabo:
This looks like a big change but it's just embedding everything in a `ZStack` and moving the `SidebarView` to the bottom.

BenGWeeks:
I assume the profile and Settings `NavigationLinks` work still? As have to be careful, `NavigationLinks`are still inside of the `NavigationView`.

joelklabo:
I didn't notice any issue with the navigation links but I'll check again, thanks!

BenGWeeks:
Try enabling all the hidden `NavigationLinks` - you get an error - it turns out Apple set a max of 10?! We will need to resolve that longer term.

jb55:
awesome, this was slightly triggering

jb55:
profile and settings buttons don't work anymore...

joelklabo:
Ah, I'll check that out

joelklabo:
Going to close this, need to figure out another strategy

PR #313, opened 2023-01-12
Fix for banner image bug

jj0b:
This PR fixes the bug where the user's banner image would be shown on other user profiles.


jj0b:
@jb55 this fixes the bug you found with banner images.

jb55:
thanks!

PR #315, opened 2023-01-13
Resolves the issue on Roboash not appearing at the top left corner + …

scoder1747:
Resolves the issue on Robohash picture not appearing at the top left corner + Also works fine with users having thier profile pic url set.

Previously "if-condition" was used because the Robohash images were not being rendered properly in some cases, but this issue does not seem to appear anymore now.

Simulator Screen Shot - iPhone 14 Pro - 2023-01-13 at 15 57 45

Simulator Screen Shot - iPhone 14 Pro - 2023-01-13 at 16 05 15

**Before the fix was applied:**
Simulator Screen Shot - iPhone 14 Pro - 2023-01-13 at 16 03 39


jb55:
thanks! this is in the merge queue.

PR #319, opened 2023-01-14
Tweak icons size on TabBar

hewigovens:
Before vs After

 <==> 

Slightly increase `Search` icon size


jb55:
not sure I like this

hewigovens:
@jb55 @bitcoinuser please take another look, the screenshot is updated

BenGWeeks:
I think this can be closed as another PR for updated icons anyway.

PR #321, opened 2023-01-14
Fix string bugs

tyiu:


damus/es-419.lproj/Localizable.stringsdict
```diff
@@ -2,6 +2,102 @@
 <!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
 <plist version="1.0">
 <dict>
+	<key>collapsed_event_view_other_notes</key>
```
tyiu, line 5: Everything here is just a placeholder until the real translation comes in.

damus/ContentView.swift
```diff
@@ -172,13 +172,20 @@ struct ContentView: View {
...
             ToolbarItem(placement: .principal) {
-                if selected_timeline == .home {
+                switch selected_timeline {
+                case .home:
                     Image("damus-home")
                     .resizable()
                     .frame(width:30,height:30)
                     .shadow(color: Color("DamusPurple"), radius: 2)
-                } else {
-                    Text("Global")
```
tyiu, line 181: This was causing `Global` to show up, even in `DM` or `Notifications` view.

damus/en-US.lproj/Localizable.stringsdict
```diff
@@ -13,9 +13,15 @@
 			<key>NSStringFormatValueTypeKey</key>
 			<string>d</string>
 			<key>zero</key>
-			<string>0 other notes</string>
+			<string>%d other notes</string>
 			<key>one</key>
-			<string>1 other note</string>
+			<string>%d other note</string>
+			<key>two</key>
```
tyiu, line 19: These are not issues for English, but other languages have different pluralization rules when the count is `two`, `few`, or `many`.

damus/Models/Mentions.swift
```diff
@@ -198,11 +198,7 @@ enum Amount: Equatable {
             let sats = NSNumber(value: (Double(amt) / 1000.0))
             let formattedSats = numberFormatter.string(from: sats) ?? sats.stringValue
 
-            if formattedSats == numberFormatter.string(from: 1) {
-                return NSLocalizedString("\(formattedSats) sat", comment: "Amount of 1 sat.")
-            }
-
-            return NSLocalizedString("\(formattedSats) sats", comment: "Amount of sats.")
+            return String(format: NSLocalizedString("sats_count", comment: "Amount of sats."), sats.decimalValue as NSDecimalNumber, formattedSats)
```
tyiu, line 201: Added one more fix here so that the string for sats formatting is consolidated into one string (rather than broken into separate code branches) so that it works for non-English languages with different pluralization rules. I'm not exactly sure, but I believe the plural rules for fractions fit under either `many` or `other`.

damus/en-US.lproj/Localizable.stringsdict
tyiu: Never mind. These get stripped out on export for English. I think Transifex knows how to add the rules for other languages. I'll revert this.

damus/en-US.lproj/Localizable.stringsdict
```diff
@@ -12,10 +12,8 @@
 			<string>NSStringPluralRuleType</string>
 			<key>NSStringFormatValueTypeKey</key>
 			<string>d</string>
-			<key>zero</key>
```
tyiu, line 15: `zero` is optional in English, and we can omit it if the string is the same as `other`.

tyiu:
OK, I'm done mucking around with this PR. Should be ready for review.

jb55:
LGTM, thx

PR #322, opened 2023-01-14
Show images reposted by yourself or people you follow

ericholguin:
Currently, when someone you follow reposts a note with an image, that image is blurred. The same is true when you repost a note with an image from someone you don't follow, that image will also be blurred to you.

This PR changes that so anytime someone you follow reposts a note with an image, that image will no longer be blurred, the thought being that you trust the people you follow.


jb55:
good idea! thx.

tclementdev:
@jb55 I think this issue has regressed, reposts from people I follow are blurred.

PR #323, opened 2023-01-14
click pfp in side menu should open profile as well

radixrat:
- I got used to clicking the pfp to open profile, now when side menu opens you click your pfp it closes the side menu, instead lets have it open the profile view.


jb55:
thx!

PR #325, opened 2023-01-15
Corrected the plurality of Reposts, Reactions, Tips, Followers, Relays

scoder1747:
Corrected the plurality of words for Reposts, Reactions, Tips, Followers, and Relays based on their count.

**Sample example for Repost:**
Simulator Screen Shot - iPhone 14 Pro - 2023-01-14 at 23 06 40

### DIFF
--- a/damus/Util/Utility.swift
+++ b/damus/Util/Utility.swift
@@ -1,0 +1,31 @@
+//
+//  Utility.swift
+//  damus
+//
+//  Created by Swift on 1/14/23.
+//
+
+import SwiftUI
+
+extension String {
+
+    /// Conform plurality of word based on the count
+    /// - Parameter count: Generic type count
+    /// - Returns: String appended with 's' if necessary
+    func conformPlurality<T>(count: T) -> String {
+        var string = self
+        if let value = count as? Int,
+            value != 1 {
+            string.append("s")
+        } else if let stringValue = count as? String,
+                  let value = Int(stringValue),
+                    value != 1 {
+            string.append("s")
+        }
+        return string
+    }
+
+    func conformPlurality<T>(count: T) -> LocalizedStringKey {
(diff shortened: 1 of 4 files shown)


damus/Views/ActionBar/EventDetailBar.swift
```diff
@@ -17,15 +17,15 @@ struct EventDetailBar: View {
             if bar.boosts > 0 {
                 Text("\(bar.boosts)")
                     .font(.body.bold())
-                Text("Reposts")
+                Text("Repost".conformPlurality(count: bar.boosts))
```
tyiu, line 20: I've already fixed this issue and it's in master:
https://github.com/damus-io/damus/pull/320/files#diff-548f04cd57f663aa1e8b4ddc8e21af44ea6ae1f2c360a0a82b8135aa5ca94fcfR1

Sorry that you spent time on this!

scoder1747:
These issues have been addressed in the Localizable strings PR, therefore changes not required.

damus/Util/Utility.swift
```diff
@@ -0,0 +1,31 @@
...
+
+extension String {
+
+    /// Conform plurality of word based on the count
+    /// - Parameter count: Generic type count
+    /// - Returns: String appended with 's' if necessary
+    func conformPlurality<T>(count: T) -> String {
+        var string = self
+        if let value = count as? Int,
+            value != 1 {
```
tyiu, line 18: This isn't the right fix. It works for only languages that have the same plurality rules as English.

Here's how I fixed it:
https://github.com/damus-io/damus/blob/cbc3c46c9d554c8829dc7ca9a9ffac48ffbcc3a1/damus/en-US.lproj/Localizable.stringsdict

The docs from Apple describe this in more detail:
https://developer.apple.com/documentation/xcode/localizing-strings-that-contain-plurals

PR #327, opened 2023-01-15
Added nsec key copying option at logout and Removed logout option from Settings

scoder1747:
A number of users complained that they forgot to copy nsec key before logging out and _they lost their accounts_.

Added copying nsec key option at logout and Remove logout option from Settings

**Demo screenshots:**
Simulator Screen Shot - iPhone 14 - 2023-02-06 at 16 50 00


damus/Views/SideMenuView.swift
```diff
@@ -158,6 +158,9 @@ struct SideMenuView: View {
                 isSidebarVisible.toggle()
             }
             .alert("Logout", isPresented: $confirm_logout) {
```
tyiu, line 160: It looks like it's recommended to use an action sheet in this case:

https://developer.apple.com/design/human-interface-guidelines/components/presentation/action-sheets

damus/Views/SideMenuView.swift
```diff
@@ -158,6 +158,9 @@ struct SideMenuView: View {
...
+                Button(NSLocalizedString("Copy nsec key", comment: "Copy nsec account key.")) {
+                    UIPasteboard.general.string = damus_state.keypair.privkey_bech32
+                }
                 Button(NSLocalizedString("Cancel", comment: "Cancel out of logging out the user.")) {
```
tyiu, line 164: >Place the Cancel button at the bottom of the action sheet

And

>Make destructive choices visually prominent. Use the destructive style for buttons that perform destructive actions, and place these buttons at the top of the action sheet where they tend to be most noticeable. For developer guidance, see destructive (SwiftUI)

Per Apple's docs

tyiu (commented):
Nice idea! I left some suggestions to conform to Apple's design guidelines.

damus/Views/SideMenuView.swift
tyiu: I think it might be this?
https://developer.apple.com/documentation/swiftui/view/confirmationdialog(_:ispresented:titlevisibility:actions:message:)-2s7pz

jb55:
checking this out now

jb55:
I think we need to integrate this into the faceid changes we made in 84ad0e03d08cfe463bb3490c6de453e59d9c627c

PR #328, opened 2023-01-15
Fix Jack's padding issue

BenGWeeks:
A simple resolution to Jack's padding issue at the top of the homepage.

There is now no gap here (between `Divider` and the timeline):

image


jb55:
thanks!

PR #329, opened 2023-01-16
name & @username in side menu open profile

olheadinternet:
added navigation links to the name and @username to open the profile view on the pull out menu of SideMenuView

was gonna do pfp as well but radixratbeat me to it. :-)


jb55:
thanks. I simplified it a bit

PR #330, opened 2023-01-16
Icons for hardclick menu event view

olheadinternet:
I was constantly confused having the two tags for user ID and note ID
so i changed the icons to clear up the confusion
the feedback i received from others was the j square version (right) is the preferable icon for copy JSON
if i did this correct you should see two commits so you can choose if you prefer the magnifying glass (left)

Screenshot 2023-01-15 162237


BenGWeeks:
Well done :-)

jb55:
thx!

PR #310, opened 2023-01-11
Fix large file crash, add SVG support, add Base64 image support

OlegAba:
This PR introduces the following changes:

1. Image file size will now be handled first in `CustomImageProcessor`. The `LargeImageProcessor` was being appended before, causing the file size check to happen after the default processor. A large file would crash the app as it would be loaded before downsampling could occur. Unfortunately, we will lose the compression that the default processor provided - many `GIFs` will be above the `1MB` size limit and not animate now. 

How to reproduce crash:
- Visit account `npub1krzjm6vekx7mcwtyx9zx30dshg2xukwkd0jnr0h6xd3hp5njzmxsc4r5ru`
- Wait for profile pic to load

2. Added support for SVG image processing. Not sure if we want to add a dependency for this feature. Let me know if you want this removed.

Before             |  After
:-------------------------:|:-------------------------:
 IMG_4568 | IMG_4602

3. Added support for Base64 png/jpg images

Before             |  After
:-------------------------:|:-------------------------:
IMG_4604 | IMG_4603

4. Moved KingFisher logic to a separate file. Not sure if this an optional design. Any feedback is welcome.

PR #335, opened 2023-01-16
Add custom image downloader - reject responses with large content length

OlegAba:
Ref: #310 conversation

Added a `20MB` content length limit for all image files


OlegAba:
Large PFP and Banner Image(172MB):
`npub1yqqwsyt595pwtgf7sax7kgfl2z8fpewc0jmjaajnv6hgfxhckutq7ydzdd`

Before             |  After
 |

damus/Models/KFImageModel.swift
```diff
@@ -106,3 +108,28 @@ struct CustomCacheSerializer: CacheSerializer {
         return DefaultCacheSerializer.default.image(with: data, options: options)
     }
 }
+
+class CustomSessionDelegate: SessionDelegate {
```
jb55, line 112: I'm confused as to where the HEAD request is happening here

OlegAba: No separate request is being made. The HEAD is being intercepted in the urlSession response which runs before the download starts in KF. I think this is more efficient and cleaner than running a separate HEAD request and then passing the url to KF.

jb55: interesting...

OlegAba: Let me know if you have any other questions or prefer it to be done another way. Also, looks like there are merge conflicts now. I can resolve those once everything looks good.

jb55: I'm just curious how you know this header is available before the data is downloaded, is there docs for this somewhere?

jb55: ah awesome

OlegAba:
@jb55 Sorry for adding so many changes for a simple rebase but I think it was needed.  The code is much cleaner now. 

Not sure why the tests are failing. It appears to be unrelated to the changes?

jb55:
the option setter thing is long overdue, thank you

PR #336, opened 2023-01-17
Lightning Zaps: native, verifiable lightning notes on nostr

jb55:
Added initial lightning zaps/tipping integration

## Todo

- [ ] Create zap request invoices within the client (zap button)
- [ ] Event zap lists
- [ ] Tally zaps on events
- [ ] Show zaps on your profile
- [ ] Anonymous zaps

Changelog-Added: Receive Lightning Zaps

### DIFF
--- a/damus/Util/Zap.swift
+++ b/damus/Util/Zap.swift
@@ -1,0 +1,322 @@
+//
+//  Zap.swift
+//  damus
+//
+//  Created by William Casarin on 2023-01-15.
+//
+
+import Foundation
+
+enum ZapSource {
+    case author(String)
+    // TODO: anonymous
+    //case anonymous
+}
+
+public struct NoteZapTarget: Equatable {
+    public let note_id: String
+    public let author: String
+}
+
+public enum ZapTarget: Equatable {
+    case profile(String)
+    case note(NoteZapTarget)
+    
+    public static func note(id: String, author: String) -> ZapTarget {
+        return .note(NoteZapTarget(note_id: id, author: author))
+    }
+    
+    var pubkey: String {
+        switch self {
+        case .profile(let pk):
+            return pk
+        case .note(let note_target):
+            return note_target.author
+        }
+    }
+    
+    var id: String {
+        switch self {
+        case .note(let note_target):
+            return note_target.note_id
+        case .profile(let pk):
+            return pk
+        }
+    }
+}
+
+struct ZapRequest {
+    let ev: NostrEvent
+}
+
+struct Zap {
+    public let event: NostrEvent
+    public let invoice: ZapInvoice
+    public let zapper: String /// zap authorizer
+    public let target: ZapTarget
+    public let request: ZapRequest
+    
+    public static func from_zap_event(zap_ev: NostrEvent, zapper: String) -> Zap? {
+        /// Make sure that we only create a zap event if it is authorized by the profile or event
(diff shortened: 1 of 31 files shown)


spcxta:
So how do we send our application/nostr metadata to the LNURL endpoint? :-P

/@jb55?amount=1&nostr=encodeURIComponent(JSON.stringify(event)) ?

jb55:
I still need to write it up, but I've been doing:

```
nostril --kind 9734 --sec $DAMUS_SEC --content "zappin you some sats!" --tags '[["p","32e1827635450ebb3c5a7d12c1f8e7b2b514439ac10a67eef3d9fd9c5c68e245"],["relays","wss://nostr-relay.wlvs.space","wss://nostr-pub.wellorder.net","wss://relay.nostr.bg","wss://nostr.v0l.io","wss://nostr.orangepill.dev"]]' > zapreq.json
```

```
curl -s "https://sendsats.lol/@jb55?nostr=$(<req.json jq -cr @uri)&amount=50000"
```

damus will be able to do this soon (zap button)

damus/Util/Zap.swift
```diff
@@ -0,0 +1,307 @@
...
+        return nil
+    }
+    
+    let zappable = payreq.allowsNostr ?? false
+    
+    var query = [URLQueryItem(name: "amount", value: "\(amount)")]
+    
+    if zappable {
+        if let json = encode_json(zapreq) {
+            query.append(URLQueryItem(name: "nostr", value: json))
```
v0l, line 284: This will result in a different flow than what is expected for LNURL, afaik the description hash must match what the `payService` responds with, in which case we would have to return the tip note there

damus/Util/Zap.swift
```diff
@@ -0,0 +1,307 @@
...
+            return nil
+        }
+        guard let bolt11 = decode_bolt11(bolt11_str) else {
+            return nil
+        }
+        /// Any amount invoices are not allowed
+        guard let zap_invoice = invoice_to_zap_invoice(bolt11) else {
+            return nil
+        }
+        guard let preimage = event_tag(zap_ev, name: "preimage") else {
```
v0l, line 74: I think the preimage is not very useful, it also prevents some people supporting this because they may not always have access to the preimage, that is the case with strike.army also

jb55: I can remove this check, you're mostly trusting the zapper anyways to determine if something is paid.

damus/Util/Zap.swift
jb55: I think its fine, wallets will not be implementing this. I think it only makes sense with nostr clients.

damus/Util/Zap.swift
jb55: I'm curious why the node sending the zap wouldn't have access to the preimage, usually it would be reacting to a payment when it sends the zap note...

jb55:
Will be changing this to remove the extraneous lnurl metadata object. will just commit directly to the 9734 note.

joelklabo:
I wanted to try and run this to see how it's gonna work but Xcode is complaining it can't find a couple types `Zap` and `ZapTarget`. You probably have your own branch but I'd like to test it if possible!

jb55:
@joelklabo should be fixed now

jb55:
merged but lots more to do

PR #337, opened 2023-01-17
Avoid showing wrong alert message when trying to re-post from Pubkey user

scoder1747:
Avoid showing wrong alert message when trying to re-post from Pubkey user.
Simulator Screen Shot - iPhone 14 Pro - 2023-01-16 at 21 16 43

### DIFF
--- a/damus/Models/DamusState.swift
+++ b/damus/Models/DamusState.swift
@@ -22,6 +22,11 @@ struct DamusState {
     var pubkey: String {
         return keypair.pubkey
     }
+
+    var is_privkey_user: Bool {
+        keypair.privkey != nil
+    }
+
     
     static var empty: DamusState {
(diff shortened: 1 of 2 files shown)


tyiu:
I actually like the alert as is, in case I accidentally press the repost button when scrolling. But that's just my opinion.

scoder1747:
> I actually like the alert as is, in case I accidentally press the repost button when scrolling. But that's just my opinion.

The pubkey logged-in user does not have the option to repost a post (as it is something like read only privilege), and right now the repost does not happen even though the alert message is shown. So showing the alert message is somewhat deceiving to the user.

tyiu:
> > I actually like the alert as is, in case I accidentally press the repost button when scrolling. But that's just my opinion.
> 
> The pubkey logged-in user does not have the option to repost a post (as it is something like read only privilege), and right now the repost does not happen even though the alert message is shown. So showing the alert message is somewhat deceiving to the user.

Oh! Sorry, I misunderstood the description. Carry on.

jb55:
uhh we should just remove the button, this is just confusing.

jb55:
oh yeah I guess so

jb55:
LGTM

PR #339, opened 2023-01-17
Fix tab views moving up after selecting from search result

OlegAba:
Before             |  After
 |


OlegAba:
Don't merge yet. Found a bug

jb55:
thanks!

PR #345, opened 2023-01-19
Removing Participants on a Thread Reply

joelklabo:
Add a participants view to choose who not to notify on a post.

The screen is launched by tapping the replied to names at the top.

**Updated:**
Simulator Screen Recording - iPhone 14 - 2023-01-19 at 16 23 41

**Behavior when all participants are removed, not positive this is right:** (the "Really cool!" note had all participants removed)
image

### DIFF
--- a/damus/Views/ParicipantsView.swift
+++ b/damus/Views/ParicipantsView.swift
@@ -11,66 +11,50 @@ struct ParticipantsView: View {
     
     let damus: DamusState
     
-    @Binding var participants: [ReferencedId]
-    @Binding var originalParticipants: [ReferencedId]
+    @Binding var references: [ReferencedId]
+    @Binding var originalReferences: [ReferencedId]
     
     var body: some View {
         VStack {
             Text("Edit participants")
             HStack {
                 Button {
-                    participants = []
+                    // Remove all "p" refs, keep "e" refs
+                    references = originalReferences.eRefs
                 } label: {
                     Text("Remove all")
                 }
                 Button {
-                    participants = originalParticipants
+                    references = originalReferences
                 } label: {
                     Text("Add all")
                 }
             }
-            ForEach(originalParticipants) { participant in
+            ForEach(originalReferences.pRefs) { participant in
                 HStack {
                     let pk = participant.ref_id
                     let prof = damus.profiles.lookup(id: pk)
                     Text(Profile.displayName(profile: prof, pubkey: pk))
                     Spacer()
                     Image(systemName: "checkmark.circle.fill")
-                        .foregroundColor(participants.contains(participant) ? .purple : .gray)
+                        .foregroundColor(references.contains(participant) ? .purple : .gray)
                 }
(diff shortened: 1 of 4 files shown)


olheadinternet:
a button to deselect all (except originalParticipant) would be nice. I tried briefly to add one to the vStack and it only removed the names after Replying To: not sure how to implement it into the HStack below to turn off all the checkmarks.

joelklabo:
> a button to deselect all (except originalParticipant) would be nice. I tried briefly to add one to the vStack and it only removed the names after Replying To: not sure how to implement it into the HStack below to turn off all the checkmarks.

Just added. I don't think my design skills are very good but I think this is probably close to a good solution

joelklabo:
Like @olheadinternet said, I still need to add remove all (except originalParticipant) too

jb55:
this looks good to me. just going to make a few tweaks and then merge

jb55:
actually there is something wrong with the logic but I've tweaked the UI so will merge this into a dev branch for now

jb55:
I tinkered with the design on the damus/remove-participants branch: 

IMG_6702

but it seems it is completely broken when making the replies, not sure why yet.

joelklabo:
> I tinkered with the design on the damus/remove-participants branch: 
> 
> 
> 
> IMG_6702
> 
> 
> 
> but it seems it is completely broken when making the replies, not sure why yet.

Yeah, sorry I tried to warn in the description it wasn't working as expected yet. Should have bolded that. 

I haven't had a chance to see why passing the 'RefenceId' is different than the old logic.

joelklabo:
Just added some fixes to this. One issue was that the `references` also contain `e` references for events as well as `p` references for people. So there were extra names at the top that just referred to keys of the events. Also, when we did remove all it would remove the references to the events too so that's probably why it was showing up as a thread starter instead of a reply @jb55

joelklabo:
Updated the design to match yours. See the new gif in the description. It looks like it working too

jb55:
will test this again in a bit

jb55:
looks good!

damus/Views/ParicipantsView.swift
```diff
@@ -0,0 +1,79 @@
...
+struct ParticipantsView: View {
+    
+    let damus_state: DamusState
+    
+    @Binding var references: [ReferencedId]
+    @Binding var originalReferences: [ReferencedId]
+    
+    var body: some View {
+        VStack {
+            Text("Edit participants")
```
tyiu, line 19: @joelklabo Since you're a core contributor and you might be submitting changes with strings frequently, just wanted to note for next time that you should add comments to localized strings so that the translators know how to approach translations. No big deal, I'm going to add comments in a separate PR. And I'm going to see if I can add a CI step to catch strings with no comments to fail the build.

joelklabo: Yeah, I realized that after it got merged. Thanks!

PR #347, opened 2023-01-19
Tagging users within Post and Reply

scoder1747:
This change allows the user to tag his Following users within Posts and Replies. This change is based on the existing Global search feature.

**Demo video below:**

https://user-images.githubusercontent.com/120697811/213552090-b98aceaf-4984-4962-a52f-c7c643b575ba.mov


joelklabo:
Nice!

olheadinternet:
nice work, not sure about the follow buttons in the list, could see accidental presses happening while selecting from the list.

scoder1747:
> nice work, not sure about the follow buttons in the list, could see accidental presses happening while selecting from the list.

I have used the existing Global search feature to populate the list with making very minimal change so that the existing flow doesn’t break out because of my change. The follow button can be removed once this is feature is stable. Thanks you!

cc @jb55

jb55:
> I have used the existing Global search feature to populate the list with making very minimal change so that the existing flow doesn’t break out because of my change

ah sorry this is definitely not the way to do this. We should be pulling from either our contact list and/or cached profiles

scoder1747:
I see what you’re saying. I think you are trying to minimize the burden to
the relays here when trying to populate the list dynamically?. But when
done in this way would allow us in the future to search the ongoing
hashtags and use them in our post/reply. I at least think that this change
would be awesome as many users are desperately waiting for this feature
(burdensome to copy/paste pubkey for tagging) until new and efficient
solutions are brought up, do you agree? Thanks

jb55:
I will try to fix this up with my suggested changed...

scoder1747:
Resolved conflicts.

jb55:
looking at this today

PR #349, opened 2023-01-20
ImageView - V2

OlegAba:
This PR significantly improves the UI/UX for the `ImageView` which had many bugs since the last update. Feel free to change the design but I think the wrapped scrollview should be used instead of creating custom gestures. If someone can please test this on a Mac as I don't have an M1/2. 

Before             |  After
 | 

Light             |  Dark
IMG_4676 | IMG_4675


jb55:
this is really good

PR #357, opened 2023-01-20
Center PFP in Zoom View

joelklabo:
Changes:
- Centered images
- Updated `x` style

| Before      | After |
| ----------- | ----------- |
|   Simulator Screen Shot - iPhone 14 - 2023-01-20 at 14 22 26 | Simulator Screen Shot - iPhone 14 Pro - 2023-01-20 at 17 02 59 |

### DIFF
--- a/damus/Views/ProfileZoomView.swift
+++ b/damus/Views/ProfileZoomView.swift
@@ -68,9 +68,9 @@ struct ProfileZoomView: View {
             .zIndex(1)
 
             VStack(alignment: .center) {
-                Spacer()
-                    .frame(height: 120)
 
+                Spacer()
+                
                 ProfilePicView(pubkey: pubkey, size: 200.0, highlight: .none, profiles: profiles)
                     .padding(100)
                     .scaledToFit()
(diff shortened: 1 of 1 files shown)


scoder1747 (approved):
This looks good. I think we also need to change the X icon making it look less bulkier?

joelklabo:
> This looks good. I think we also need to change the X icon making it look less bulkier?

Yeah, that's a good idea

joelklabo:
I think that looks better @scoder1747 tried to match Twitter.

damus/Views/ProfileZoomView.swift
```diff
@@ -54,23 +54,19 @@ struct ProfileZoomView: View {
...
-                }
+            Button {
+                presentationMode.wrappedValue.dismiss()
+            } label: {
+                Image(systemName: "xmark")
+                    .foregroundColor(.white)
+                    .font(.subheadline)
+                    .padding(.leading, 20)
             }
-            .zIndex(1)
```
scoder1747, line 68: Changes look good. But we can keep the Z index as we still want the X mark to be shown in front of the image while we zoom in to the fullest!

scoder1747: Also  you can close the later pr?

joelklabo: Oh yeah, good call. I'll fix that and close the other 👍

jb55:
thx!

PR #359, opened 2023-01-21
Fix localized string for privacy access description for photos

tyiu:
This came at the suggestion of one of our translators for Austrian German.


damus Localizations/en-US.xcloc/Localized Contents/en-US.xliff
```diff
@@ -285,6 +285,16 @@ Number of profiles a user is following.</note>
         <target>DM</target>
         <note>Navigation title for DM view, which is the English abbreviation for Direct Message.</note>
       </trans-unit>
+      <trans-unit id="DM Type" xml:space="preserve">
```
tyiu, line 288: @jb55 Anytime we change strings, we need to go to Xcode and run `Product` -> `Export Localizations...` and select `English (United States)` only (no other languages should / need to be checked off) so that it can run through the translation pipeline.

I'll make this automated soon, but fyi in the meantime.

damus/Views/ConfigView.swift
```diff
@@ -91,7 +91,7 @@ struct ConfigView: View {
                     Section(NSLocalizedString("Secret Account Login Key", comment: "Section title for user's secret account login key.")) {
                         HStack {
                             if show_privkey == false {
-                                SecureField(NSLocalizedString("PrivateKey", comment: "Title of the secure field that holds the user's private key."), text: $privkey)
```
tyiu, line 94: Unrelated change but this was caught by one of the translators. This seems to be a typo.

damus Localizations/en-US.xcloc/Localized Contents/en-US.xliff
jb55: hmm ok good to know

PR #360, opened 2023-01-21
Update CI Build to Specify Xcode Version

joelklabo:
Test were often failing because the Xcode version was not specified and the simulator and iOS version we test on was not always available on the version it chose. This also updated the name to be a little more descriptive.


tyiu (approved):
Nice, I was wondering why CI was failing for my pull request.

joelklabo:
Yeah, it kept trying to run on different versions of Xcode and then I couldn't find the simulator.

jb55:
thanks!

PR #366, opened 2023-01-22
fix for seeing 2 "add post" buttons when swiping

thomasRademaker:
#### Note `this PR also fixes a broken test`
move the postButtonContainer view to avoid having 2 buttons on screen when swiping tabs

Before 

https://user-images.githubusercontent.com/1885323/213920452-439e91c1-b78e-4b5b-a1e9-dfe22f911ee1.mov

After

https://user-images.githubusercontent.com/1885323/213920461-199dbcc3-a62e-4528-8413-9e0127b7ffee.mov

### DIFF
--- a/damus/ContentView.swift
+++ b/damus/ContentView.swift
@@ -93,15 +93,23 @@ struct ContentView: View {
 
     var PostingTimelineView: some View {
         VStack {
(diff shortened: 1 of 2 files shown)


jb55:
awesome, thanks!

PR #371, opened 2023-01-23
Make text in logout warning a NSLocalized string

realprogrammersusevim:
The text in an alert about saving your private key before logging out was plain text and not a localized string.

### DIFF
--- a/damus/Views/SideMenuView.swift
+++ b/damus/Views/SideMenuView.swift
@@ -165,7 +165,7 @@ struct SideMenuView: View {
                     notify(.logout, ())
                 }
             } message: {
-                Text("Make sure your nsec account key is saved before you logout or you will lose access to this account", comment: "Reminder message in alert to get customer to verify that their private security account key is saved saved before logging out.")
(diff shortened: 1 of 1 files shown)


damus/Views/SideMenuView.swift
```diff
@@ -165,7 +165,7 @@ struct SideMenuView: View {
                     notify(.logout, ())
                 }
             } message: {
-                Text("Make sure your nsec account key is saved before you logout or you will lose access to this account", comment: "Reminder message in alert to get customer to verify that their private security account key is saved saved before logging out.")
+                Text(NSLocalizedString("Make sure your nsec account key is saved before you logout or you will lose access to this account", comment: "Reminder message in alert to get customer to verify that their private security account key is saved saved before logging out."))
```
tyiu, line 168: Is this necessary? I thought that strings in Text components are automatically deemed as a localized string. That's how I left it for other parts of code and it seemed fine.

realprogrammersusevim: Good catch. You're completely correct https://developer.apple.com/documentation/swiftui/localizedstringkey. I need to read some more documentation.

PR #376, opened 2023-01-23
Add Reposts view

tyiu:
Copied mostly from the Reactions view. Somewhat related, I noticed that sometimes the Reposts counter is incorrect and doesn't match with the list (the list seems more accurate), as demonstrated in the animated gif:

Simulator Screen Recording - iPhone 14 Pro - 2023-01-22 at 23 22 47

### DIFF
--- a/damus.xcodeproj/project.pbxproj
+++ b/damus.xcodeproj/project.pbxproj
@@ -12,6 +12,9 @@
 		3169CAED294FCCFC00EE4006 /* Constants.swift in Sources */ = {isa = PBXBuildFile; fileRef = 3169CAEC294FCCFC00EE4006 /* Constants.swift */; };
 		31D2E847295218AF006D67F8 /* Shimmer.swift in Sources */ = {isa = PBXBuildFile; fileRef = 31D2E846295218AF006D67F8 /* Shimmer.swift */; };
 		3A4325A82961E11400BFCD9D /* Localizable.stringsdict in Resources */ = {isa = PBXBuildFile; fileRef = 3A4325AA2961E11400BFCD9D /* Localizable.stringsdict */; };
+		3AA247FD297E3CFF0090C62D /* RepostsModel.swift in Sources */ = {isa = PBXBuildFile; fileRef = 3AA247FC297E3CFF0090C62D /* RepostsModel.swift */; };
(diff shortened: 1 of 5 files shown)


tyiu:
> Was just checking why this test failed. Doesn't seem like it's related to your change: `damusUITests-Runner (6463) encountered an error (Early unexpected exit, operation never finished bootstrapping - no restart will be attempted. (Underlying Error: Test crashed with signal kill before starting test execution. If you believe this error represents a bug, please attach the result bundle at`
> 
> I was already thinking it would be worth disabling the UITests b/c they take a long time to load (GitHub gives us 2000 minutes a month for free) and we don't have any real ones. This failed in a UITest. I think I'll open a PR doing that. Should speed things up at least if/until we have some.

@joelklabo Might be worth looking into other CI solutions. I haven't used it, but CircleCI provides 30,000 credits per month, which I think is 6,000 build minutes per month, for open source projects that need to build on macOS.

joelklabo:
> > Was just checking why this test failed. Doesn't seem like it's related to your change: `damusUITests-Runner (6463) encountered an error (Early unexpected exit, operation never finished bootstrapping - no restart will be attempted. (Underlying Error: Test crashed with signal kill before starting test execution. If you believe this error represents a bug, please attach the result bundle at`
> > I was already thinking it would be worth disabling the UITests b/c they take a long time to load (GitHub gives us 2000 minutes a month for free) and we don't have any real ones. This failed in a UITest. I think I'll open a PR doing that. Should speed things up at least if/until we have some.
> 
> @joelklabo Might be worth looking into other CI solutions. I haven't used it, but CircleCI provides 30,000 credits per month, which I think is 6,000 build minutes per month, for free for open source projects that need to build on macOS. https://circleci.com/pricing/#comparison-table

Wow, yeah that's pretty good. I'll take a look.

joelklabo:
Actually looks like the 2000 is for private repositories so maybe it's more here too:

image

tyiu:
> Actually looks like the 2000 is for private repositories so maybe it's more here too:
> 
> image

If you scroll down to `Frequently asked questions` underneath `What if I am building open source?`, it says `Open source Windows and macOS builds or private repositories can use the 30,000 credits offered in our free plan. If you are building a bigger open source project and would like the flexibility of our new plans, let us know`

jb55:
awesome, thx

PR #377, opened 2023-01-23
Disable UI Tests

joelklabo:
Disabling UI Tests. They take a long time to run and right now they are just the example ones that come with a new iOS project. Github gives 2000 minutes of CI build time per month for free so this should extend that and make the builds faster. Also less false failures.


jb55:
2b832120ec494a9f3052d01828b8a0acdb77a1bd

PR #378, opened 2023-01-23
Remove redundant logout button in config view

realprogrammersusevim:
Since there's now an easy to access button to logout of Damus on the side bar I didn't see a need for another logout button in the config view and removed it.


jb55:
thanks! merging soon

PR #380, opened 2023-01-23
Fix relay counter in fr translation

pedrohiiq:
If you change language to french, the counter for the relays will be broken, this should fix it.


tyiu:
@pedrohiiq Thanks for helping fix! Indeed, merging this in will fix the issue. But we're managing translations here: https://explore.transifex.com/damus/damus-ios/

That means that any changes you make manually in these files will get overridden by the Transifex integration. If you would like to contribute in translation issues, please join the Transifex team above.

For this specific issue, I've already fixed it in Transifex. The integration just hasn't pushed it yet since it's waiting for all 100% of strings to be translated.

tyiu:
Closing in favor of https://github.com/damus-io/damus/pull/392/files#diff-6204e00d789ca0fdc9a6802df1e5acda95216620de48c15cffaff233ca501286R78

See my comment above.

PR #383, opened 2023-01-23
Fix to allow relays using ws:// protocol

### LINKED ISSUE #344: Can no longer add insecure relay
This [commit](https://github.com/damus-io/damus/commit/b688fa98a5498ebd7d27e59d8709af10e7b88ce2#diff-a38dacb1a8a41057dff5e3331305ea3d5cf329d2cdb9f5734c663a05850b3477R141-R144) introduces a bug where we can no longer add insecure relays. Ie. those relays using the protocol: `ws://`.

I enter: `ws://192.168.0.10`, and this gets added as: `wss://ws://192.168.0.10`.

This line:
```
if relay.starts(with: "wss://") == false {
```

Needs to be changed to something like:
```
if relay.starts(with: "wss://") == false && relay.starts(with: "ws://") == false {
```

Keep up the awesome work!

nevets963:
Closes #344

### DIFF
--- a/damus/Views/ConfigView.swift
+++ b/damus/Views/ConfigView.swift
@@ -154,7 +154,7 @@ struct ConfigView: View {
                     return
                 }
                 
-                if relay.starts(with: "wss://") == false {
+                if relay.starts(with: "wss://") == false && relay.starts(with: "ws://") == false {
                     relay = "wss://" + relay
                 }
                 


jb55:
oh whoops. thanks for catching this.

jb55:
thanks! 1baae90bebb9bb1fab8f05377aa0e480f1cdf0f1

PR #384, opened 2023-01-23
Bump pfp/banner animated fize size limit to 5MiB/20MiB

jb55:
Since users are unhappy.

cc @OlegAba 

Changelog-Changed: Bump pfp/banner animated fize size limit to 5MiB/20MiB


OlegAba:
Sounds good. I'm looking into why some valid images are failing to load.

PR #391, opened 2023-01-24
Custom Picker - Posts, Posts & Replies and DMs, Requests

ericholguin:
This PR replaces the Segmented Picker that was being used for the Filters Views and the DirectMessagesView with a Custom Picker, which implements the same functionality but with a custom Damus style. Matches that one Bird App.


jb55:
oh nice, I like this

PR #393, opened 2023-01-25
Added In before component - displays when a user first created an event

ericholguin:
This PR adds the "In before" component to the ProfileView. It is displaying the month and year of the user's first event, since there is no real way of knowing when a user joined nostr. The name "In before" is technically accurate since the user joined before that event and also has been widely posted on nostr as a meme at this point.

### DIFF
--- a/damus/Components/InBefore.swift
+++ b/damus/Components/InBefore.swift
@@ -1,0 +1,41 @@
+//
+//  InBefore.swift
+//  damus
+//
+//  Created by eric on 1/24/23.
+//
+
+import SwiftUI
+
+struct InBefore: View {
+    @Binding var events: [NostrEvent]
+
+    var body: some View {
+        HStack {
+            Image(systemName: "calendar")
+                .foregroundColor(.gray)
+                .font(.footnote)
+            
+            Text(in_before_text(events: events))
+                    .font(.footnote)
+                    .foregroundColor(.gray)
+        }
+    }
+    
+    func in_before_text(events: [NostrEvent]) -> String {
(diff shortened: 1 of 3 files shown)


jb55:
this is probably not doing what you think it is. this wouldn't show you the user's oldest event. also created_at dates could be faked.

ericholguin:
Yeah my assumption was that events are ordered by the 'created_at' key, so the last event is assumed to be the first created. I'm also guessing this might be an issue due to relays being used and it might not grab the actual oldest event. I'll dig into the code more but if you can share how it works i'd appreciate it. As for created_date being faked, can NIP 22 solve for this?

damus/Components/InBefore.swift
```diff
@@ -0,0 +1,41 @@
...
+    
+    func in_before_text(events: [NostrEvent]) -> String {
+        let dateFormatter = DateFormatter()
+        let currentDate = Int64(Date().timeIntervalSince1970)
+        let oldestEvent = Date(timeIntervalSince1970: TimeInterval(events.last?.created_at ?? currentDate))
+        dateFormatter.dateFormat = "LLLL"
+        let monthString = dateFormatter.string(from: oldestEvent)
+        dateFormatter.dateFormat = "yyyy"
+        let yearString = dateFormatter.string(from: oldestEvent)
+        return "In before \(monthString) \(yearString)"
```
tyiu, line 33: You'll need to wrap this with NSLocalizedString and add a comment so that it can be translated to other languages.

ericholguin:
Closing - no solution for this yet and not critical

PR #397, opened 2023-01-25
ImageView improvements

OlegAba:
Double tap gesture

 

Content inset

Before             |  After
 | 

Safe area bug

Before             |  After
 | 

Swipe to dismiss bug

Before             |  After
 |


jb55:
thx!

PR #398, opened 2023-01-26
Use AttributedString in NoteContentView

joelklabo:
This is similar to my previous `AttributedString` PR but only deals the the `NoteContentView`. See old PR here: https://github.com/damus-io/damus/pull/301

There was a second `ContentParser` in that PR that was used for user profiles because they aren't parsed into `Block` like note content is. So to handle that we would need to add some other parser like the `ContentParser` did or parse them the same way content is parsed, which might cause issues with link previews, invoices etc.

Link types tested:
- [x] Users
- [x] Notes
- [x] Hashtags
- [x] URLs (the text version that stays in the note content

### DIFF
--- a/damus/Views/NoteContentView.swift
+++ b/damus/Views/NoteContentView.swift
@@ -9,13 +9,13 @@ import SwiftUI
 import LinkPresentation
 
 struct NoteArtifacts {
-    let content: String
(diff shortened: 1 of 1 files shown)


jb55:
yay!

PR #413, opened 2023-01-27
Show EULA prior to login

tyiu:
Previously, EULA was shown only when creating an account.

Simulator Screen Recording - iPhone 14 Pro - 2023-01-27 at 00 17 26


scoder1747 (commented):
I have run this branch and it works! but is it possible to persist the `accepted` state such that if the user logs out and logs in back, he wont be shown the EULA page? tq

jb55:
thanks!

PR #416, opened 2023-01-27
Fix SVG processing crash

OlegAba:
The app would crash when converting SVG data with no size to UIImage. Unfortunately, SVGKit does not provide a way to catch errors so I replaced the package with CoreSVG - a private Apple framework used for SFSymbols. 

Ref: https://gist.github.com/ollieatkinson/eb87a82fcb5500d5561fed8b0900a9f7

Testing:
Loading the following profile will result in a crash...
`npub1ea6y6er4n7vt9t5cxdf2p2jhp2ppga5texfmrma0xqj6t6frudaszaw8g5 `


jb55:
thanks!

PR #418, opened 2023-01-28
npub QR Code

ericholguin:
This PR adds the QRCode view that links to an npub.

Adds a QRCode button to the side menu which presents a QR Code fullscreen and can be dismissed once it has been used.

qrcode


jb55:
I fixed this up and I have added it to the merge queue. thx!

PR #425, opened 2023-01-29
Fix bug with copying user pubkey

tyiu:
Without fix, it copies your own pubkey instead of the OP's.


scoder1747 (approved):
This works. Screenshot:
Simulator Screen Shot - iPhone 14 Pro - 2023-01-28 at 22 51 20

PR #430, opened 2023-01-29
Add export and import translation scripts

tyiu:
Makes it slightly easier to maintain translations. Separately, we should take this a step further by integrating it into CI.


jb55:
👍

PR #431, opened 2023-01-30
Fix author attribution to the actual translators in CHANGELOG

tyiu:
Transifex integration doesn't propagate this information to GitHub unfortunately.


jb55:
thanks!

PR #433, opened 2023-01-30
Add fa-IR localization

mahdi:
This PR adds Persian (fa-IR) localization.


tyiu:
@mahdi Appreciate your efforts here. We're actually managing translations over at https://explore.transifex.com/damus/damus-ios/

Could you please join there and add the translations through the UI instead? Once they're 100% translated, there's an integration that automatically creates a pull request.

mahdi:
> @mahdi Appreciate your efforts here. We're actually managing translations over at https://explore.transifex.com/damus/damus-ios/
> 
> Could you please join there and add the translations through the UI instead? Once they're 100% translated, there's an integration that automatically creates a pull request.

Sure! Thanks @tyiu!

PR #434, opened 2023-01-30
Fix height of DM input

tyiu:
There's still a bug where the scroll view doesn't shift up proportionally to the height of the text editor input, but I couldn't figure that one out.

Before:
Simulator Screen Recording - iPhone 14 Pro - 2023-01-30 at 12 02 56

After:
Simulator Screen Recording - iPhone 14 Pro - 2023-01-30 at 12 01 58


jb55:
thx!

PR #455, opened 2023-01-31
Add nostr-pub.semisol.dev as a bootstrap relay

Semisol:
Performance is good from testing. Resources aren't being maxed out (yet.)

Hosted with Hetzner in the EU


jb55:
is this a free relay?

PR #456, opened 2023-01-31
remove landscape orientation for iOS

jacany:
Landscape orientation is pointless on iOS, and gets in the way when laying down.

This pull request removes the landscape orientation just for iOS, this doesn't affect iPadOS.


jb55:
for some reason some people want this

PR #479, opened 2023-02-01
Relay Detail View

joelklabo:
Adds a new view to show relay details (NIP-11), also links to supported NIPs on GitHub

*Updated Design*

| Light      | Dark |
| ----------- | ----------- |
| light |  dark |

### DIFF
--- a/damus/Views/Relays/RelayDetailView.swift
+++ b/damus/Views/Relays/RelayDetailView.swift
@@ -11,7 +11,7 @@ import WrappingHStack
 struct RelayDetailView: View {
     let relay: String
     
-    @State private var networkError = false
+    @State private var networkError: String?
     @State private var nip11: RelayNIP11?
 
     var body: some View {
@@ -34,8 +34,8 @@ struct RelayDetailView: View {
                     }
                 }
                 .padding()
-            } else if networkError {
-                Text("error")
+            } else if let networkError {
+                Text(networkError)
                     .foregroundColor(.red)
             } else {
                 ProgressView()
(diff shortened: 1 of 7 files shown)


jb55:
i missed this!

joelklabo:
Yeah, I like it!

jb55:
I seem to get "failed to parse nip11 data" on most relays

joelklabo:
Hmm, I'll check it out

joelklabo:
Ok this should be working now. I had an extra field in there that wasn't actually required. And I added a default empty value so if something is missing it still works.

For the Damus relay and the one with the yellow connection there is an actual error. Damus seems to be down and the other times out so those make sense:

Simulator Screen Recording - iPhone 14 Pro - 2023-02-04 at 14 00 13

jb55:
will try this again

damus/Views/Relays/RelayDetailView.swift
```diff
@@ -0,0 +1,113 @@
...
+            } else if let errorString {
+                Text(errorString)
+                    .foregroundColor(.red)
+                    .font(.caption)
+            } else {
+                ProgressView()
+            }
+        }
+        .navigationTitle(nip11?.name ?? "")
+        .onAppear {
```
jb55, line 62: you can do `.task` instead of `onAppear` and then use the async version of URLSession.shared...

jb55: I'll make this change don't worry about it

joelklabo: Oh nice, didn't know that. Thanks!

PR #481, opened 2023-02-02
Fix text truncation in CarouselItemView

hanton:
Text in CarouselItemView would be truncated in small screen devices, eg. iPhone SE. 
This PR sets `minimumScaleFactor` to  fit text into available space, following UI style in [ProfileView/EditButton](https://github.com/damus-io/damus/blob/e40cc9a50a06ac9cecf7cecbd96ef5790b4d76e6/damus/Views/ProfileView.swift#L94). 

Before             |  After
Before  |  After


jb55:
thanks!

hanton:
It appeared when I opened Damus first time in my phone, so put up the PR for it ;)

Thanks for the project!

PR #484, opened 2023-02-02
Add Valera relays to bootstrap list

617a7a:
Following up on #471, here is Valera's relay network. These are anycasted on our server infrastructure, and run [nostr-rs-relay](https://git.sr.ht/~gheartsfield/nostr-rs-relay).


jb55:
nostr-rs-relay can't really handle the load for our users unfortunately

617a7a:
the endpoint is load balanced across multiple datacenters - it's not a single relay :)

do you recommend any other implementations?

Issue #480, opened 2023-02-02
Add What's a Relay Explanation [UX][Newbies]


alltheseas:
## User Story ##

As a newbie user who is learning about Damus and nostr, I want to want to know "what are those numbers in the upper corner", so I can reduce the uncertainty of this new experience.

### Acceptance Criteria ###

1. Relay `x/y` is clickable ✅ (https://github.com/damus-io/damus/pull/491 completed criteria 1.)
<img width="409" alt="Screenshot 2023-02-01 at 9 01 26 PM" src="https://user-images.githubusercontent.com/64376233/216220963-fd13165f-e5d7-483c-99b6-5b2f3ac50f25.png">

2. Click leads to the following screen (in relay config. See #491)
<img width="338" alt="Screenshot 2023-02-01 at 9 01 07 PM" src="https://user-images.githubusercontent.com/64376233/216220984-8fdc6cc5-3e42-49e9-bd70-7357aa5225e4.png">

3. `Got it` exits the modal
4. `Learn more` leads to https://nostr-resources.com/

User feedback:
![image](https://user-images.githubusercontent.com/64376233/216672337-964da309-6146-4461-b711-eda3401e362a.png)

Origin: rock@cash.app AKA @rockymedure
npub1yauhh489aefdkzsewe5vj2u68elzxlsln7nn5yxr34e3c22vljdqkn7h9m

https://nostr.build/i/nostr.build_6ce088636bd065aafcbc59a76f9ff5853212ba741d540fee0504a8325a89fb67.gif

https://www.figma.com/file/SCeX7GoDI7LTI03KTIAza3/Relay-Education?node-id=0%3A1

https://damus.io/note1gq6m92zakw6k59n35wy509jl9c09a0mgdmrar8sd4x04xpr33m7q3vhv9d

radixrat:
If I get time tonight I’ll see if I can try a few things out and post up if no one else tackles it, I think relay confusion is a big problem though so it would be good to educate on it in the relay config page (or have a separate resource to learn more)

SNQ-2001:
I want to work on this issue!

radixrat:
Didn’t get time to look at it, please go ahead ✌️

devandsev:
[...]

PR #491, opened 2023-02-02
clicking on relay numbers on home view brings you to config

radixrat:
- clicking on relay numbers on home view brings you to relay config 

Noticed something up with dutch localization I had to remove it to test my code so think something is broken there in main branch.

### DIFF
--- a/damus/ContentView.swift
+++ b/damus/ContentView.swift
@@ -260,9 +260,11 @@ struct ContentView: View {
                                     ToolbarItem(placement: .navigationBarTrailing) {
                                         HStack(alignment: .center) {
                                             if home.signal.signal != home.signal.max_signal {
-                                                Text("\(home.signal.signal)/\(home.signal.max_signal)", comment: "Fraction of how many of the user's relay servers that are operational.")
-                                                    .font(.callout)
-                                                    .foregroundColor(.gray)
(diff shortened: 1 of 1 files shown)


radixrat:
> @radixrat you interested in this ticket related to clicking on the relay numbers: #480

Do you think we could put the explainer on the relayconfig view instead as some sort of question mark button? I would rather have the relays connected / relays link on home go to the config view personally instead of just informational modal.

radixrat:
> > _I don't know if there is enough surface area to do it all from the timeline view, if they are curious about it they will click it and then we could give an explainer at that point I would think (on the actual page with relays)_
> 
> If user clicks on relay numbers on timeline view, can you take them directly to the relay page?
> 
> > _I also don't know what @jb55 policy is on linking to remote sites so we would have to discuss that as well._
> 
> If no guidance, can you remove the "`Learn More`" button for time being.

Yes, thats what this PR does, when you click 13/15 it will bring them to the relay config page where we could put a describer modal similar to what you were trying to do on the home timeline

jb55:
> Noticed something up with dutch localization I had to remove it to test my code so think something is broken there in main branch.

fixed this

jb55:
looks good! thx

PR #492, opened 2023-02-02
Blur profile pics if not following with some exceptions

realprogrammersusevim:
Damus blurs images from people you don't follow but profile pics are a huge loophole. None of them are blurred. In the NoteContentView, if the image is blurred than the profile pic should be blurred as well.

I did make some exceptions. Obviously the users profile pic is never blurred but it's also not blurred in the profile view. This could be subject to change but my reasoning was that the profile pic should be blurred when just scrolling and the event isn't from a friend, but if you choose to visit the profile than all bets are off.


radixrat:
I have a similar PR that reverts to robohash if they aren't in your load remote image policy range.  I know blur looks better but my personal opinion is we shouldn't be loading images that we are just going to blur anyways (especially for people outside of friend circle).  Couldn't get much traction with it though so maybe people like the blur better: https://github.com/damus-io/damus/pull/127

realprogrammersusevim:
> I have a similar PR that reverts to robohash if they aren't in your load remote image policy range. I know blur looks better but my personal opinion is we shouldn't be loading images that we are just going to blur anyways (especially for people outside of friend circle). Couldn't get much traction with it though so maybe people like the blur better: #127

I don't think the PRs are incompatible at all. Blur simply shows that you don't follow the person. My PR will still blur robohashes that you don't follow even though robohashes aren't NSFW at all. Sometimes people you follow don't have a pfp or it isn't loading. This PR will help distinguish those people because their robohash will be clear.

jb55:
I think this would make user experience much worse. Maybe there's a better way

realprogrammersusevim:
> I think this would make user experience much worse. Maybe there's a better way

What exactly would make it worse? Isn't this the same as blurring images from people you don't follow? I would guess that the user experience would be far worse if a new user opens the app, checks the global feed, and sees crystal clear hardcore porn gifs as profile pictures.

jb55:
I think we need to make this a configurable option, on by default. Same with blurred pfps. @radixrat is your PR up to date?

realprogrammersusevim:
> I think we need to make this a configurable option, on by default. Same with blurred pictures. @radixrat is your PR up to date?

Good idea. I'll add a toggle to the config settings for blurring profile pics set to on by default.

radixrat:
> I think we need to make this a configurable option, on by default. Same with blurred pictures. @radixrat is your PR up to date?

yes https://github.com/damus-io/damus/pull/127 ive been keeping it up to date because I use it on my local copy

realprogrammersusevim:
Added a configuration option for blurring profile pictures that is on by default. Made the section name in the config "Images" to make room for any additional config options that @radixrat would like to add. Is this ready to merge now @jb55?

jb55:
now that we have radix's change, do we need this?

realprogrammersusevim:
Yes, the default is still loading all images like before, which means profile picture blurring is still useful. Even if the user decides to change the image loading settings, blurring improves user experience by distinguishing the friends without a profile picture and and other people who's profile pictures haven't been loaded.

jb55:
I noticed images are not blurred on global when you have "everyone" image policy enabled. I don't like this. I will probably revert @radixrat's change for now.

radixrat:
> I noticed images are not blurred on global when you have "everyone" image policy enabled. I don't like this. I will probably revert @radixrat's change for now.

The point was to give clients choice over what content they wanted to load to their phone, blurring images means you are still downloading porn on your phone from the network perspective, this is why friend of friend was the default.

realprogrammersusevim:
> I noticed images are not blurred on global when you have "everyone" image policy enabled. I don't like this. I will probably revert @radixrat's change for now.

Why not separate those options so image loading and blurring are separate settings?

jb55:
> > I noticed images are not blurred on global when you have "everyone" image policy enabled. I don't like this. I will probably revert @radixrat's change for now.
> 
> The point was to give clients choice over what content they wanted to load to their phone, blurring images means you are still downloading porn on your phone from the network perspective (people use their phones on work networks and get fired, etc), this is why friend of friend was the default.

doing friend-of-friend kills pfp loading everywhere, which seems extreme when explicit PFPs aren't that common.

radixrat:
> > > I noticed images are not blurred on global when you have "everyone" image policy enabled. I don't like this. I will probably revert @radixrat's change for now.
> > 
> > 
> > The point was to give clients choice over what content they wanted to load to their phone, blurring images means you are still downloading porn on your phone from the network perspective (people use their phones on work networks and get fired, etc), this is why friend of friend was the default.
> 
> doing friend-of-friend kills pfp loading everywhere, which seems extreme when explicit PFPs aren't that common.

Friend of friend makes pfp only load from friends of friends, IP harvesting can be done from any image hosted externally

jb55:
looks like this bitrotted

PR #494, opened 2023-02-02
Allow copying text in bio

suhailsaqan:
It's often useful to be able to copy text from a person's bio such as pubkeys of other accounts.

Simulator Screen Shot - iPhone 14 Pro - 2023-02-02 at 14 19 09


suhailsaqan:
@alltheseas yea I can pick that up.

jb55:
thanks!

PR #496, opened 2023-02-03
Switch between Posts and Posts & Replies in Profile View

ericholguin:
This PR adds the selector for posts and posts & replies to the profile view, allowing users to view posts only or posts and replies that the user has posted.

profile-timeline-selection


jb55:
yayyy

jb55:
the scroll should probably be on everything. will have to refactor these views

ericholguin:
> the scroll should probably be on everything. will have to refactor these views

Fixed it

jb55:
sorry for the delay, trying this again

PR #500, opened 2023-02-03
Fix profile edit button text to not wrap

tyiu:
Occurs when accessibility text settings for font size are set to the smallest.

Before:
Simulator Screen Shot - iPhone SE (3rd generation) - 2023-02-03 at 00 02 18

After:
Simulator Screen Shot - iPhone SE (3rd generation) - 2023-02-03 at 00 01 39


jb55:
thx

PR #502, opened 2023-02-03
Fix PostView screen freeze when posting long note

aki-mizu:
Currently when user tries to send a long enough note that takes time to get response from relays, the PostView screen freezes until note was sent successfully. 
This creates a bad user experience that user might tap post button again when PostView screen is frozen, which takes even longer time plus duplicate content will be sent again.
This pull request tries to solve this problem by closing PostView immediately after tapping post button.
The 0.3 seconds delay is needed because it takes roughly 0.3 seconds for the PostView to disappear completely.


jb55:
hmm ok... but why is it blocking? It should be instant. seems weird...

aki-mizu:
As far as I know because NSNotificationCenter is synchronous. It must wait for the observer's handler to complete before the next line of execution can happen. That’s why the dismiss method won’t be called unless onReceive(handle_notify(.post) finishes send event causing blocking of PostView screen.

jb55:
hmm ok i'll test and merge this soon

aki-mizu:
A new commit has been added to make it easier to call send_post function. It should work the same as before.

jb55:
should be fixed now

PR #504, opened 2023-02-03
Update ContentView.swift

Barondane:
This should fix a bug preventing changing views between Posts / Posts & Replies in AR localization and probably in other RTL languages


Barondane:
While it looked like it fixed the issue, it introduces a new issue. Will look for something better.

PR #505, opened 2023-02-03
Update Localizable.stringsdict

Barondane:
Some translations did not show up in the app and they were in English in this file. Replaced the English strings here to AR.


tyiu:
Closing as I'll import it using the standard process with the import scripts.

tyiu:
I see the problem. Apple has a .stringsdict format that allows us to define pluralization rules. Arabic has more complex rules "few", "many", than other languages, such as English. English doesn't have "few" and "many" pluralization rules, so when it gets sent to Transifex also probably doesn't get them for Arabic. This seems like a bug in the tooling. I'll manually pull in your changes for now.

PR #507, opened 2023-02-03
Lightweight png files

pea-sys:
Lightweighting of png files by zopflipng.
There is no change in appearance.
* 15 items
* 574847 bytes → 204211 bytes


jb55:
thank you!

jb55:
reverted this actually because it changes the visual style of the logo

PR #515, opened 2023-02-03
Export and import translations, remove de_AT in favor of de, and move zh to zh-Hans

tyiu:


jb55:
thanks! in merge queue

PR #522, opened 2023-02-04
Add DeepL translation integration

tyiu:
Simulator Screen Shot - iPhone 14 Pro - 2023-02-04 at 15 05 08


jb55:
if I were to subscribe to this, does it charge per use or something?

tyiu:
> if I were to subscribe to this, does it charge per use or something?

Only if you explicitly choose to use the Pro plan.

For the Free plan, they ask for credit card (unfortunate) to guarantee that you're a unique user and not abusing free API limits with creating multiple accounts. Once you exceed the limit, the calls will fail with a 4xx response.

tyiu:
> > if I were to subscribe to this, does it charge per use or something?
> 
> 
> 
> Only if you explicitly choose to use the Pro plan.
> 
> 
> 
> For the Free plan, they ask for credit card (unfortunate) to guarantee that you're a unique user and not abusing free API limits with creating multiple accounts. Once you exceed the limit, the calls will fail with a 4xx response.

Also, to be clear, the Free plan gives 500k characters per month. Pro plan is $5.49 USD + $25.00 per 1 million characters.

jb55:
very cool. tested and works. merging

PR #530, opened 2023-02-05
add accessibility labels to EventActionBar

bryanmontz:
Added the first few accessibility labels to Damus. These labels help iOS' VoiceOver to know how to communicate to vision-impaired users.

Prior to this change, in VoiceOver mode, the device would say "Button" for all of the buttons on the EventActionBar, except for the like button, which would say "shaka-line" (to a user who has no idea what that means). After this change, it will say "Share button", "Like button", etc.


jb55:
oh good to know, thanks

PR #533, opened 2023-02-05
Fix Unit Tests

joelklabo:
- Missing `contacts` parameter in DM unit tests
- Remove 1000 year ago test (this crosses into Julian calendar which probably isn't worth supporting 🤣)


joelklabo:
FYI @jb55 @tyiu

jb55:
thanks!

PR #541, opened 2023-02-06
Fix UX to open relay config view when navigating from personal profile

tyiu:
Simulator Screen Recording - iPhone 14 Pro - 2023-02-06 at 10 58 17


joelklabo:
I was thinking I should add the detail view there but this is even better! #479

PR #552, opened 2023-02-08
Fix determine_is_request logic

hanton:
Based on description in [DirectMessagesView.swift](https://github.com/damus-io/damus/blob/0563ec8bf86585a7bc0246c25af1d1af0c992b52/damus/Views/DirectMessagesView.swift#L70): 

```
Text("DMs", comment: "Picker option for DM selector for seeing only DMs that have been responded to. DM is the English abbreviation for Direct Message.")
    .tag(DMType.friend)
Text("Requests", comment: "Picker option for DM selector for seeing only message requests (DMs that someone else sent the user which has not been responded to yet). DM is the English abbreviation for Direct Message.")
    .tag(DMType.rando)
```

The return value of `determine_is_request` should be: 
`true`: if all of the events(messages) are coming from the user(`our_pubkey`)
`false`: if one of the events is not coming from the user

### DIFF
--- a/damus/Models/DirectMessageModel.swift
+++ b/damus/Models/DirectMessageModel.swift
@@ -19,7 +19,7 @@ class DirectMessageModel: ObservableObject {
     
     func determine_is_request() -> Bool {
         for event in events {
-            if event.pubkey == our_pubkey {
+            if event.pubkey != our_pubkey {
                 return false
             }
         }


jb55:
This doesn't seem right. The goal of this was if any of the DMs is ours then we've replied, so it is not a request. If we have not replied then it is a request.

hanton:
You are right. Thanks for the clarification. The issue I'm trying to solve is https://github.com/damus-io/damus/issues/556. 

Will close this PR and dig into it more.

Issue #539, opened 2023-02-06
Username autocompletion with @ should be case insensitive


jblachly:
Currently, when composing a note, searching for users to autocomplete from your contact list via '@' does a case sensitive search. This typically works fine since most users have used lowercase handles. However, if a user has an uppercase character, Damus fails to match.

Test vectors:
`@nvk` (no results)
`@NVK` (matches pubkey)

jblachly:
Correct; I was not specific enough in my original issue report; updated the text to reflect this. Thanks!

PR #553, opened 2023-02-08
Make autocompletion case insensitive

larrysalibra:
This is my attempt at fixing #539.

**The problem:**

Before this commit, if I am trying to mention @bitkarrot on Nostr, autocompletion doesn't work because her name is capitalized on her Nostr profile.

**The solution**

After this commit, search strings and names are converted to lowercase when looking for matches so that typing "@bitkarrot" also suggests (capital B) "Bitkarrot"

**Behavior before:**

IMG_3056
IMG_3057

**Behavior after:**

Simulator Screen Shot - iPhone 14 Pro Max - 2023-02-08 at 15 55 41
Simulator Screen Shot - iPhone 14 Pro Max - 2023-02-08 at 15 55 48

### DIFF
--- a/damus/Views/Posting/UserSearch.swift
+++ b/damus/Views/Posting/UserSearch.swift
@@ -77,7 +77,7 @@ func search_users(profiles: Profiles, tags: [[String]], search: String) -> [Sear
         
         let profile = profiles.lookup(id: pubkey)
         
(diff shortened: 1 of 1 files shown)


larrysalibra:
I see this is already fixed in master. Withdrawing this PR. Sorry about the noise! 🙇

Issue #556, opened 2023-02-08
New created 1:1 conversation goes to "Requests" list, instead of "DMs" list


hanton:
Repro steps:
1. Navigate to `Search` tab
2. Tap on a user profile image to navigate to the user profile view
3. Tap on the `Message` button to navigate to message view
4. Send a message
5. Tap on the `Message` tab
6. The newly created conversation is in the "Requests" list, instead of "DMs" list

video:

https://user-images.githubusercontent.com/2510864/217678380-630ac1d0-0d6e-4f40-a775-3c3570460989.mov

PR #558, opened 2023-02-09
Fix empty our_pubkey in DirectMessagesModel

hanton:
The PR is for solving https://github.com/damus-io/damus/issues/556

Root cause:
1. `dms` is created with empty pub key in [HomeModel](https://github.com/damus-io/damus/blob/9ab03034a2ffc6abfd454f165598d768a3e99db6/damus/Models/HomeModel.swift#L58)
2. After that, its pub key is never updated even when `damus_state` is updated in [ContentView](https://github.com/damus-io/damus/blob/9ab03034a2ffc6abfd454f165598d768a3e99db6/damus/ContentView.swift#L581)
3. The empty pub key of `DirectMessagesModel` is used to init a new [DirectMessageModel](https://github.com/damus-io/damus/blob/9ab03034a2ffc6abfd454f165598d768a3e99db6/damus/Models/DirectMessagesModel.swift#L32), so the pub key of newly created `DirectMessageModel` is also empty
4. As a result, `determine_is_request` in [DirectMessageModel](https://github.com/damus-io/damus/blob/9ab03034a2ffc6abfd454f165598d768a3e99db6/damus/Models/DirectMessageModel.swift#L22) will always return `true` because `our_pubkey` is empty. 
5. It causes newly created conversations show in the `Requests` list, instead of `DMs` list

Input is welcome.

### DIFF
--- a/damus/Models/DirectMessagesModel.swift
+++ b/damus/Models/DirectMessagesModel.swift
@@ -10,7 +10,7 @@ import Foundation
 class DirectMessagesModel: ObservableObject {
     @Published var dms: [(String, DirectMessageModel)] = []
     @Published var loading: Bool = false
-    let our_pubkey: String
+    var our_pubkey: String
     
     init(our_pubkey: String) {
         self.our_pubkey = our_pubkey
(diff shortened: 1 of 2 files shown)


jb55:
ohhh

jb55:
sorry will get back to this soon

PR #560, opened 2023-02-09
Fix side menu offset

OlegAba:
Fixes the side menu protruding when in the horizontal hidden state

IMG_4811

### DIFF
--- a/damus/Views/SideMenuView.swift
+++ b/damus/Views/SideMenuView.swift
@@ -18,6 +18,7 @@ struct SideMenuView: View {
     
     var sideBarWidth = min(UIScreen.main.bounds.size.width * 0.65, 400.0)
     let verticalSpacing: CGFloat = 20
+    let padding: CGFloat = 30
     
     func fillColor() -> Color {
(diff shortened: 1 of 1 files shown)


BenGWeeks:
This build fails for me.

CleanShot 2023-05-28 at 16 28 48

Not sure why that doesn't work as not related to your code change. To get to work, just changed:

```
                ForEach(artifacts.links, id:\.self) { link in
                    //if let url = link {
                        //LinkViewRepresentable(meta: .url(url))
                        LinkViewRepresentable(meta: .url(link))
                            .frame(height: 50)
                    //}
                }
```

But other than that, this seems to fix the left side bar when in landscape :-)

FYI @jb55

jb55:
On Sun, May 28, 2023 at 08:36:24AM -0700, Ben Weeks wrote:
>But other than that, this seems to fix the left side bar when in landscape :-)
>
>FYI @jb55

Thanks for the ping! Merged in ff151562975ff92366cb3191d8884a9792c9fe71

PR #561, opened 2023-02-09
Minor Post View improvements

ericholguin:
This PR adds some minor view improvements to the Post View.  You can now see your profile image while posting and the post button has been damus-ified.


jb55:
is there any way to make the pfp not move when you start typing?

PR #562, opened 2023-02-09
Remove trailing slash when adding a relay

middlingphys:
Remove trailing slash from relay URI to be added to
- prevent relays with duplicate URIs from being registered
- unify the appearance of URIs in the relay list


jb55:
good call

Issue #511, opened 2023-02-03
Screen-size / -resolution issue - iPhone SE (A1723) - Unable to join/register


gitro1983:
Hi, 

I've got an old iPhone SE here (A1723), EULA agreement shows everything and I can see what I'd agree to and validate.

After validating I only can see a part of the content on the screen and also unable to scroll. (Could probably validate again but I can't see the full key "Account-ID npub..." at the bottom, above "parts of" the "Create-button" -> makes no sense to continue..)

Screen info:  
    Retina 
    4-inch (diagonal) widescreen LCD Multi-Touch display
    1136‑by‑640‑pixel resolution at 326 ppi

G.

jarbus:
I have the 2020 SE, and I literally can't press the "create an account" button because of this issue.

OlegAba:
@gitro1983 @jarbus

If I understand correctly, the keyboard is blocking the create button? You can hide the keyboard by tapping outside of the fields or tapping the return button. See video below. We will work on making this page scrollable to not confuse to new users.

https://user-images.githubusercontent.com/19398259/217937342-f15a51ce-3236-4f9b-ab62-0cd9aa678e4d.mov

gitro1983:
Hi Oleg,

no, this has nothing to do with the iphone onscreen-keyboard, this is a general layout / screensize-resolution/-size issue, (img1= screen after start app (look at that image/gif/jpg for ex.), img2= create account-screen) (I didn't send the EULA agreement-screen now (because it works fine)) look: 

![IMG_3402](https://user-images.githubusercontent.com/124392630/217943617-402101c0-ba5a-46db-9ac2-a81a515927e0.PNG)
![IMG_3404](https://user-images.githubusercontent.com/124392630/217943623-6cbf6ce3-dd36-4b31-8ac5-1e75ab0a3c03.PNG)
(input fields not fully displayed, npub not fully displayed+not scrollable e.t.c..)

G.

jarbus:
The create button isn't even on the screen for me

![image](https://user-images.githubusercontent.com/42819002/217948512-00027642-5d3c-4133-a93b-5241a605b979.png)

gitro1983:
It's probably better to look after jarbus screen-issue first probably.
I've noticed right now, I've missed an IOS update.. updating IOS right now..sorry.

OlegAba:
@gitro1983 

This is a bug with translation. Will fix it soon. What language are you using?

OlegAba:
[...]

PR #566, opened 2023-02-10
Restrict dynamic font type - max font size

OlegAba:
Issue ref: #511 

The UI completely breaks on some views when large accessibility font is used. This change will limit font size in the entire app to be no bigger than xxxLarge. Eventually, the layout should be redesigned to support the accessibility sizes.

Using Accessibility 5 size:
Before             |  After
IMG_4839 | IMG_4840

### DIFF
--- a/damus/damusApp.swift
+++ b/damus/damusApp.swift
@@ -32,6 +32,7 @@ struct MainView: View {
                     }
             }
         }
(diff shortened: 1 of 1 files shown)


jb55:
thanks!

jb55:
merged in 0bdec912f814ecc11bffad2c0b9fff7e5a17b066

PR #567, opened 2023-02-11
Don't Show Follows You on Your Own Profile

joelklabo:


jb55:
could probably just put this logic in the follows_you var?

joelklabo:
> could probably just put this logic in the follows_you var?

Good idea, I'll try that

joelklabo:
Yeah, that's much better @jb55

jb55:
oh I only meant the view-local var. this logic is incorrect now since you can actually follow yourself.

joelklabo:
I see, so maybe there was never a bug in the first place. I saw follows you on my profile. But I guess I actually do

PR #568, opened 2023-02-11
Add menu ellipsis button to notes

ericholguin:
This PR replaces the context menu (long press on notes) with a menu button, shown as ellipsis, on each note. This will allow for other long press gestures on notes while still allowing users to access the context menu. 
menu-ellipsis

### DIFF
--- a/damus/Views/ChatroomView.swift
+++ b/damus/Views/ChatroomView.swift
@@ -24,7 +24,7 @@ struct ChatroomView: View {
                                  next_ev: ind == count-1 ? nil : thread.events[ind+1],
                                  damus_state: damus
                         )
-                        .event_context_menu(ev, keypair: damus.keypair, target_pubkey: ev.pubkey)
+                        .contextMenu{MenuItems(event: ev, keypair: damus.keypair, target_pubkey: ev.pubkey)}
                         .onTapGesture {
                             if thread.initial_event.id == ev.id {
                                 //dismiss()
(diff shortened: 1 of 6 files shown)


tyiu:
Love it. This has been bothering me for a while.

jb55:
I actually did this awhile back but couldn't figure out how to do a single-click instead of long click on the ellipses

jb55:
somehow this works though. amazing.

jb55:
this is so good. thank you

jb55:
Had to revert this since it's causing that navigation issue

OlegAba:
@ericholguin @jb55 

I don't believe the changes in this PR are directly causing the navigation bug. For some reason, this bug shows up simply by removing the existing `TextEvent` context menu in Master.
Remove `line 69` in [TextEvent (Master)](https://github.com/damus-io/damus/blob/master/damus/Views/Events/TextEvent.swift) to reproduce the bug:
`.event_context_menu(event, keypair: damus.keypair, target_pubkey: pubkey, bookmarks: damus.bookmarks)`

image-carousel-bug-1

The cause of the bug seems to be the `ImageCarousel` context menu. I am unable to reproduce the bug after removing `line 213-217` in [ImageCarousel (Master)](https://github.com/damus-io/damus/blob/master/damus/Components/ImageCarousel.swift)
```
.contextMenu {
     Button(NSLocalizedString("Copy Image", comment: "Context menu option to copy an image to clipboard.")) {
         UIPasteboard.general.string = url.absoluteString
     }
}
```
</br>

Is the outer context menu masking the inner context menu bug? I'm not sure why this happening or how to fix it other than removing the inner context menu. This might be a SwiftUI bug. Maybe we should remove the inner context menu until we can resolve the issue? The only menu option for the `ImageCarousel` is copying the image url which is also available in the image viewer. Let me know what you all think.

jb55:
thanks! merged in b6a7f52596aff4208e526a9bc3c3a65b591f01ad

PR #573, opened 2023-02-12
Add replies count to selected event action bar

tyiu:
Replies count shows up only when you tap into an event as the threaded information is only fetched when you select the event. The count and icon will turn blue if you have replied.

Simulator Screen Shot - iPhone 14 Pro - 2023-02-12 at 02 52 57 MediumSimulator Screen Shot - iPhone 14 Pro - 2023-02-12 at 02 53 08 MediumSimulator Screen Shot - iPhone 14 Pro - 2023-02-12 at 02 53 21 Medium


jb55:
going to use our eventcounter for this instead

PR #575, opened 2023-02-12
Strip punctuation from URLs

### LINKED ISSUE #574: Punctuation part of URL
Common punctuation is considered part of URLs, which results in non-existing URLs:
Screenshot 2023-02-12 at 13 55 26

Only when a URL is directly followed by a whitespace, will the URL work as intended.

Proposed solution: common punctuation should not be considered part of the ending of a URL.

eval:
2 fixes:
- slight adjustment to `consume_until_whitespace`
- fix #574

Screenshot 2023-02-12 at 14 14 31


jb55:
this looks right to me, thx!

PR #576, opened 2023-02-12
improve EventActionBar button spacing

bryanmontz:
Improved the layout of the EventActionBar buttons. Before this change, they are not very aesthetically pleasing because the spacing between them isn't even.

spacing-pr


jb55:
nice!

PR #580, opened 2023-02-12
Add ability to unlike content.

newdeveloper19954:
In order to unlike content we need to add new event [[NIP-9]](https://github.com/nostr-protocol/nips/blob/master/09.md) and update app flow to track liked and unliked content properly.

ezgif com-video-to-gif


jb55:
logic looks good!

jb55:
doesn't seem to work for me

newdeveloper19954:
@jb55 will test it more today. You try regular flow like and unlike ?

PR #581, opened 2023-02-13
Delete user settings store on logout

tyiu:
Logout should probably get rid of all local settings tied to the pubkey. Some settings are currently global. If we want to have multi-account support in the future, we'd need to update user settings store to support that and then update the settings deletion to be scoped only to that.

### DIFF
--- a/damus/ContentView.swift
+++ b/damus/ContentView.swift
@@ -473,6 +473,18 @@ struct ContentView: View {
         .onReceive(handle_notify(.new_mutes)) { notif in
             home.filter_muted()
         }
+        .onReceive(handle_notify(.logout)) { _ in
+            guard damus_state != nil else {
+                return
+            }
+
+            do {
+                try damus_state!.settings.delete_settings(damus_state!.pubkey)
+            } catch {
+                // Could not delete all settings for some reason. Continue with logout.
+                print("Unable to delete all user settings for \(damus_state!.pubkey). Continuing with logout.")
+            }
+        }
         .alert(NSLocalizedString("Deleted Account", comment: "Alert message to indicate this is a deleted account"), isPresented: $is_deleted_account) {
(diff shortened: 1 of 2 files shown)


damus/Models/UserSettingsStore.swift
```diff
@@ -227,6 +227,19 @@ class UserSettingsStore: ObservableObject {
             return deepl_api_key != ""
         }
     }
+
+    func delete_settings(_ pubkey: String) throws {
```
tyiu, line 231: There's some danger here where if we introduce new settings, we need to remember to add it here. Not sure what the best way to mitigate that risk would be. But this is better than what we have currently.

damus/ContentView.swift
```diff
@@ -473,6 +473,18 @@ struct ContentView: View {
...
+        .onReceive(handle_notify(.logout)) { _ in
+            guard damus_state != nil else {
+                return
+            }
+
+            do {
+                try damus_state!.settings.delete_settings(damus_state!.pubkey)
+            } catch {
+                // Could not delete all settings for some reason. Continue with logout.
+                print("Unable to delete all user settings for \(damus_state!.pubkey). Continuing with logout.")
```
tyiu, line 485: This can technically happen when trying to delete the stored API keys for LibreTranslate and DeepL. Not sure how that could happen though. The vault deletion method throws so we have to catch it.

jb55:
why are we deleting settings on logout? I'm confused. Then all your settings would be reset when you logout!? I logout all the time to switch accounts, this seems like it would be a bad experience.

tyiu:
> why are we deleting settings on logout? I'm confused. Then all your settings would be reset when you logout!? I logout all the time to switch accounts, this seems like it would be a bad experience.

Hmm, that's fair. I think a better way would be to scope settings to the pubkey maybe. And then only allow writes if logged in with the privkey? Not sure

PR #588, opened 2023-02-13
Beautify eula

BenGWeeks:
Hi, I have now formatted the EULA so it is styled correctly and can be localised by section.

Could you please review @tyiu, as I am thinking instead of say:

```
Text(NSLocalizedString("By using our Application, you signify your acceptance of this EULA. If you do not agree to this EULA, you may not use our Application.",comment: "Acceptance of Terms paragraph in the EULA"))
```

We should use a shorter key, so if the text is to be updated in the future, you don't need to update the Key in all the localizations, e.g.

```
Text(NSLocalizedString("EULA-Acceptance-of-Terms-Paragraph",comment: "Acceptance of Terms paragraph in the EULA"))
```

### DIFF
--- a/damus/Views/EULAView.swift
+++ b/damus/Views/EULAView.swift
@@ -17,105 +17,103 @@ struct EULAView: View {
             DamusGradient()
             
             ScrollView {
-                //Text("End User License Agreement", comment: "Label indicating that the below text is the EULA, an acronym for End User License Agreement.")
-                //    .font(.title.bold())
-                //    .foregroundColor(.white)
-                
-                Divider()
+                Text(NSLocalizedString("End User License Agreement", comment: "Title of the EULA page"))
+                    .font(.title.bold())
+                    .foregroundColor(.white)
                 
                 // Introduction
                 Group {
-                    Text("Introduction")
+                    Text(NSLocalizedString("Introduction", comment:"Introduction heading in the EULA"))
                         .font(.title2)
                         .frame(maxWidth: .infinity, alignment: .leading)
                         .padding()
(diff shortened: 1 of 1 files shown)


jb55:
is this the way to do it @tyiu ?

tyiu:
Instinctively, breaking up the EULA into different parts means that translators lose some context. This feels like a single document that needs to be translated in one go, especially if it's meant to be "legally binding" or whatever. But that makes it difficult to apply the styling.

I need to sleep on this. Let me think about it.

BenGWeeks:
How about we use `Text(NSLocalizedString("EULA-Part-01"))` etc. @tyiu, so it will all be in the right order for the translators? (PS You have Ukrainian from my colleague, and Hungarian from my wife inbound :-))

tyiu:
> How about we use `Text(NSLocalizedString("EULA-Part-01"))` etc. @tyiu, so it will all be in the right order for the translators? (PS You have Ukrainian from my colleague, and Hungarian from my wife inbound :-))

Sorry for not getting back to you! It's been busy and this is a hard problem.

I still think breaking up the EULA into separate strings, regardless of how the keys are organized, make it difficult for the translators to translate it properly, especially since it's "legal" text.

I have an alternative proposal. What if we put the EULA in a separate Markdown file and send that file off for translations. We could use [Apple's Markdown parser](https://github.com/apple/swift-markdown) and convert them into SwiftUI components. I've thought about this approach for only about 5 minutes so feel free to poke holes in it. Happy to spend some time with you on it in person at Nostrica if that's not too long of a wait. I'm currently in a weird time zone and on a trip that makes it difficult for me to collaborate synchronously.

Also, yes! I've been collaborating with your Ukrainian colleague and looking forward to collaborating / meeting with your wife!

PR #612, opened 2023-02-14
Force default zap amount text field to accept only numbers

tyiu:
Inspired by https://stackoverflow.com/a/58736068


jb55:
thanks!

PR #613, opened 2023-02-14
Created translation to czech language cs-CZ

Chuck3CZ:
I made translation to **czech** 🇨🇿, only translated, not tested.
I want to test it before update to AppStore, but I don't  have access to TestFlight version. 

Can I build it on my own? Need I anything more to add this file?

Or can you add me to Testflight with version of this language?


tyiu:
Thanks for contributing to the Czech translations! We actually manage translations through Transifex. Please join here:
https://www.transifex.com/damus/settings/

I added cs (Czech) -- I took off the -CZ because it seems like Czech only has one regional variant. As far as how it's listed in Transifex anyway.

I uploaded the translations file from your pull request onto Transifex. There's a few new strings that popped up since you worked on this. Feel free to continue on Transifex.

Unfortunately, @jb55 has no more TestFlight slots to give, but he can correct me if I'm wrong. You can certainly build this on your own. As long as you have Xcode, you can build it.

You can contact me on nostr: `npub1yaul8k059377u9lsu67de7y637w4jtgeuwcmh5n7788l6xnlnrgs3tvjmf`

Chuck3CZ:
Thanks, added all translation of missing strings on Transifex.

PR #614, opened 2023-02-15
Fix frequent Kingfisher hang and other scrolling improvements

OlegAba:
This PR should, hopefully, fix the hang that many users have been experiencing related to Kingfisher
Ref: [`note13j4kv0qc6eju03yc58kqhd00tuekc68z6pz8tk8y8y55l92c5rdslf9mw8`](https://damus.io/note13j4kv0qc6eju03yc58kqhd00tuekc68z6pz8tk8y8y55l92c5rdslf9mw8)

I believe the hang is caused when many images are queued for download from scrolling  quickly; too many callbacks on the main thread at once, or network blocking? `cancelOnDisappear` will help for this. 

According to KF docs [line 98-101](https://github.com/onevcat/Kingfisher/blob/master/Sources/General/KingfisherOptionsInfo.swift), `backgroundDecode` will also help eliminate scrolling hitches. **IMPORTANT**: If used, this option will have a significant impact on memory when downloading images. Older devices will crash while loading large files, even with a downsampler. If we decide to use this option, I recommend we implement #335 first.
- The accounts [`npub1ea6y6er4n7vt9t5cxdf2p2jhp2ppga5texfmrma0xqj6t6frudaszaw8g5`](https://damus.io/npub1ea6y6er4n7vt9t5cxdf2p2jhp2ppga5texfmrma0xqj6t6frudaszaw8g5) is following will help with testing older devices. Includes PFP and Banner for popular file types in small/medium/large/xlarge sizes.

Removed `fade` and `loadDiskFileSynchronously` as it impacts performance and doesn't make a noticeable UX difference.

# Instruments Tests
Ran the following flow before and after changes

## Results Before (Contains KFImage.ImageBinder hang)
before

## Results After (Less frequent and shorter delays in hangs/hitches)
after

### DIFF
--- a/damus/Components/ImageCarousel.swift
+++ b/damus/Components/ImageCarousel.swift
@@ -231,10 +231,10 @@ struct ImageCarousel: View {
                         KFAnimatedImage(url)
                             .callbackQueue(.dispatch(.global(qos: .background)))
                             .processingQueue(.dispatch(.global(qos: .background)))
+                            .cancelOnDisappear(true)
+                            .backgroundDecode()
                             .cacheOriginalImage()
-                            .loadDiskFileSynchronously()
                             .scaleFactor(UIScreen.main.scale)
-                            .fade(duration: 0.1)
                             .configure { view in
                                 view.framePreloadCount = 3
                             }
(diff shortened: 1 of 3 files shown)


jb55:
let's get #335 rebased and I'll merge these at the same time

jb55:
this is huge, thank you. will test

jb55:
merged this! let's merge 335 next

PR #619, opened 2023-02-15
fix: ProfileView: fix clipping of "Copied" and add animation for polish

bryanmontz:
Fixed an issue where the word "Copied" gets clipped on the ProfileView when you copy a pubkey. I also added animation for polish.

Before:
before

After:
after


bryanmontz:
Agree. Done!

Before:
icon-copy-before


After:
icon-copy-after  icon-copy-after-dark

bryanmontz:
After those screenshots, I updated the circle checkmark too. It was pretty low-res as well.

iPhone_12_mini

bryanmontz:
The tests are currently broken in `master`.

jb55:
awesome! thx will test

PR #620, opened 2023-02-15
Remove untranslatable strings from export, export and import translations

tyiu:


jb55:
thx!

tyiu:
Amended commit to include Indonesian in case you haven't pulled it in yet.

jb55:
I'll pull again

PR #628, opened 2023-02-16
fixed tests

bryanmontz:


bryanmontz:
Root cause:
When a DM arrives from a new person, the `inserted` flag was not being set, which prevented the sort from happening.

jb55:
thanks! merged in 1b4e54582fa01095066cf75f8776a429eb285ed5

PR #634, opened 2023-02-17
top-level tab state restoration

bryanmontz:
Before this change:
Every time the app launches, you are shown the Home tab and you're on Posts & Replies. What if I like to primarily use the app for DMs? Or what if I like to obsessively check my Notifications tab?

After this change:
The app will restore you to whichever of the four main tabs you were on last, and for the Home tab, it will also remember if you were on Posts or Posts & Replies.

Here it is opening directly to Notifications from a cold launch:

restore

### DIFF
--- a/damus/ContentView.swift
+++ b/damus/ContentView.swift
@@ -62,7 +62,7 @@ struct ContentView: View {
     @State var status: String = "Not connected"
     @State var active_sheet: Sheets? = nil
     @State var damus_state: DamusState? = nil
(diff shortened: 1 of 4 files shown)


bryanmontz:
To test restoration in a simulator:
1. Launch the app
2. Navigate to one of the main tabs other than the default
3. Home out of the app (Cmd+Shift+H)
4. Quit the app with the stop button in Xcode

The next time the app is run, it will start on the previously selected tab.

PR #635, opened 2023-02-17
code clean-up: @discardableResult, unused params, simplifications

bryanmontz:
Minor code clean-ups:

`let _ = ...`
Use `@discardableResult` to let the compiler know you might not need the result, or check `result != nil` for easier readability with the same meaning.

Use `Array.first(where:)` and `Array.filter(:)` to select items from an array, rather than looping through, for better readability and faster lookups.


jb55:
thx

Issue #513, opened 2023-02-03
Make Post Text Selectable [Feature Request]


gregorhintner:
I suggest making post text, in post detail view, text cursor selectable, to easily allow copying out text segments. 
If possible, the margin areas should continue to respond the long presses for the context menu.

gregorhintner:
> Is this a duplicate of #493

It matches the current rewrite of #493 , though the original titles have nothing directly in common.
I will close this issue, but hope #493 retains its current scope. Thanks for notifying me.

PR #639, opened 2023-02-17
Note text selection

OlegAba:
Feature request: #513

I think we should follow Twitter's UI/UX for text selection - text is selectable only on the main/selected note. Selecting text on a view that has a gesture (NavigationLink) does not work well.

**Needs testing for Mac**

### Twitter Text Selection
https://user-images.githubusercontent.com/19398259/219731600-5f6c7ad6-62ac-4f0d-bcf6-9225893ebbe1.mov

</br>

### Damus Text Selection
https://user-images.githubusercontent.com/19398259/219736239-ee6503a6-5acf-4d53-b43f-fd1941b7372a.mov


tyiu:
Agreed about making it behave how Twitter works and this is amazing. Thank you! So if there's a url in the note, it doesn't work well?

I was looking into this yesterday and the native SwiftUI Text component doesn't allow fine-grained text selection like this, only copying and sharing the entire text, so I gave up. Looks like you had to use the UIView components to custom make it.

OlegAba:
Thanks @tyiu . There's no SwiftUI wrapper for UITextView so I think using a ViewRepresentable is the only option right now. 

Selection works the same on links. Ideally, navigation to the links should be pushed to the stack but this is how it worked before the changes. The selection is only buggy when it's used on a view with an existing gesture. I think this is why Twitter avoids it.


https://user-images.githubusercontent.com/19398259/219773825-50b41b0b-d020-4acf-b88d-a07b105096f2.mov

jb55:
so awesome. testing now

PR #644, opened 2023-02-18
refinements to RelayConnection plus tests for creating requests

bryanmontz:
Added some polish to the `RelayConnection` class:

1. Used `final class` to explicitly denote that this class is not intended to be subclassed (helps with compile time and communicates clear intent to all readers).
2. Marked all properties and functions with the tightest possible access controls, to make future debugging easier and simplify the external interface. This includes using `private(set)` to indicate a property that is read-write to the owning type and read-only to outside types.
3. Made `socket` optional to avoid wasting memory resources.
4. Used the `TimeInterval` type for times. It's an alias for `Double` and matches Apple's APIs more accurately.
5. Removed all unnecessary `self.`s for property and function accessing.
6. Made use of Swift's excellent protocol extension feature to isolate the functions that convert `NostrRequest`s into `String`s from the `RelayConnection` class itself. This also removes the functions from the global namespace to prevent Xcode from suggesting them to any type that doesn't need them. Using them now requires explicit conformance to the protocol.
7. Added tests for the same functions to demonstrate another benefit of the protocol, which the test class can also conform to for access.

### DIFF
--- a/damus/Nostr/RelayConnection.swift
+++ b/damus/Nostr/RelayConnection.swift
@@ -13,49 +13,47 @@ enum NostrConnectionEvent {
     case nostr_event(NostrResponse)
 }
 
-class RelayConnection: WebSocketDelegate {
-    var isConnected: Bool = false
-    var isConnecting: Bool = false
-    var isReconnecting: Bool = false
-    var last_connection_attempt: Double = 0
-    var socket: WebSocket
-    var handleEvent: (NostrConnectionEvent) -> ()
-    let url: URL
+final class RelayConnection: WebSocketDelegate, NostrRequestMaking {
+    private(set) var isConnected = false
+    private(set) var isConnecting = false
+    private(set) var isReconnecting = false
+    
+    private(set) var last_connection_attempt: TimeInterval = 0
+    private var socket: WebSocket?
+    private var handleEvent: (NostrConnectionEvent) -> ()
+    private let url: URL
 
     init(url: URL, handleEvent: @escaping (NostrConnectionEvent) -> ()) {
         self.url = url
         self.handleEvent = handleEvent
-        // just init, we don't actually use this one
-        self.socket = make_websocket(url: url)
     }
     
     func reconnect() {
-        if self.isConnected {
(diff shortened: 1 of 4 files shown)


jb55:
> Made socket optional to avoid wasting memory resources

don't we always need socket?

bryanmontz:
Yes, but the problem was that it was creating and throwing away a socket every time a RelayConnection was allocated. I thought of a better way that eliminates the optionality and doesn't throw away a socket. Will send another commit.

jb55:
hmm ok I see what you mean now... that commit looks better.

jb55:
need to make sure that doesn't break anything

bryanmontz:
Right. Super critical code. I'd love for it to be as clean as possible and have tests.

damus/Nostr/RelayConnection.swift
```diff
@@ -114,51 +114,47 @@ class RelayConnection: WebSocketDelegate {
...
-func make_nostr_req(_ req: NostrRequest) -> String? {
-    switch req {
-    case .subscribe(let sub):
-        return make_nostr_subscription_req(sub.filters, sub_id: sub.sub_id)
-    case .unsubscribe(let sub_id):
-        return make_nostr_unsubscribe_req(sub_id)
-    case .event(let ev):
-        return make_nostr_push_event(ev: ev)
+protocol NostrRequestMaking {}
+extension NostrRequestMaking {
```
jb55, line 120: I'm actually not a fan of this pattern. This just looks like class extensions. I'm pretty happy with pure top-level functions, don't need to get fancy here. I suggest in the future to make all of these changes separate so I can merge them individually.

bryanmontz: That's okay. I can back out that part. Wanted to see what you thought of the pattern.

bryanmontz: Done. I'll leave that out next time.

jb55:
thanks!

jb55:
seems to be working ok!

PR #648, opened 2023-02-18
Quotes

ericholguin:
This PR changes the boost/repost button from an alert to a sheet presentation, which shows the option to repost or quote a note. Repost works the same. Pressing the quote button presents the user with a post view sheet with the note id already inside.

The Relays view was also improved slightly to use the new half sheet and use the accent color on the participants text so its more obvious that it is "tap-able"

_**NOTE**_: This could be improved to actually show the embedded note and move the text cursor.

quotes


jb55:
quote note should be a new kind, not the note ref hack that people use today. I was thinking kind 66, with content as the note and then the note you're quoting is json encoded in a tag

ericholguin:
Should it be a new kind? Can't it be a kind 1 with tags? Nip 01 mentions this:

> This NIP defines "p" — meaning "pubkey", which points to a pubkey of someone that is referred to in the event —, and "e" — meaning "event", which points to the id of an event this event is quoting, replying to or referring to somehow.

ericholguin:
Closing - I need to do a refactor and take a different approach

PR #649, opened 2023-02-18
Add Bookmarking (Local to device)

joelklabo:
| Example |
| - |
| Simulator Screen Recording - iPhone 14 Pro - 2023-02-20 at 21 29 40 |

**UPDATED**
No settings button anymore, moved to navigation bar in bookmarks view

| Example (Dark) | Empty State | Empty State (Dark) | Settings | 
|      -          |      -         |      -      |       -      |
| example-dark | empty | example-dark | settings |

This just saves them (as `JSON`) to user defaults so only local to device.

### DIFF
--- a/damus/Models/BookmarksManager.swift
+++ b/damus/Models/BookmarksManager.swift
@@ -1,0 +1,40 @@
+//
+//  BookmarksManager.swift
+//  damus
+//
+//  Created by Joel Klabo on 2/18/23.
+//
+
+import Foundation
+
+class BookmarksManager {
+    static private let userDefaults = UserDefaults.standard
+    
+    static var bookmarks: [String] {
+        get {
+            return userDefaults.stringArray(forKey: "bookmarks") ?? []
+        }
+        set {
+            let uniqueBookmarks = Array(Set(newValue))
+            if uniqueBookmarks != bookmarks {
+                userDefaults.set(uniqueBookmarks, forKey: "bookmarks")
+            }
+        }
+    }
+    
+    static func isBookmarked(_ string: String) -> Bool {
+        return bookmarks.contains(string)
+    }
+    
+    static func updateBookmark(_ string: String) {
+        if isBookmarked(string) {
+            bookmarks = bookmarks.filter { $0 != string }
+        } else {
+            bookmarks.append(string)
+        }
+    }
+    
+    static func clearAll() {
+        bookmarks = []
+    }
+}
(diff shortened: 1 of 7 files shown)


jb55:
dope!

damus/Models/BookmarksManager.swift
```diff
@@ -0,0 +1,40 @@
...
+//
+
+import Foundation
+
+class BookmarksManager {
+    static private let userDefaults = UserDefaults.standard
+    
+    static var bookmarks: [String] {
+        get {
+            return userDefaults.stringArray(forKey: "bookmarks") ?? []
```
jb55, line 15: this needs to be keyed on pubkey

jb55: I have a pretty dumb helper for that: pk_setting_key

jb55 (commented):
looks good, just one change needed

damus/Models/BookmarksManager.swift
joelklabo: Oh, right, multiple accounts? I can update that 🤙

joelklabo:
Ok, updated @jb55

jb55:
looks great! testing

jb55:
just need a way to delete and then I can merge

joelklabo:
> just need a way to delete and then I can merge

Must be a bug but it should show "Un-bookmark" if you've already bookmarked it. I'll check again, would that work? If it worked?

jb55:
It shows "Bookmark".

jb55:
Also we probably shouldn't have "clear bookmarks" in settings, we can just add it as a button on the top-right of the toolbar on the bookmarks view

joelklabo:
> Also we probably shouldn't have "clear bookmarks" in settings, we can just add it as a button on the top-right of the toolbar on the bookmarks view

Yeah, that would be better. I'll try that

joelklabo:
Ok, I think this is all fixed up now, updated the GIF in the description with current behavior @jb55

joelklabo:
Looks like one more thing to figure out after rebasing with master:

`Cannot convert value of type 'Binding<[NostrEvent]>' to expected argument type 'EventHolder'`

joelklabo:
> Looks like one more thing to figure out after rebasing with master:
> 
> `Cannot convert value of type 'Binding<[NostrEvent]>' to expected argument type 'EventHolder'`

Should be fixed now

PR #650, opened 2023-02-19
refactor similar RepostsModel and ReactionsModel into one parent class

bryanmontz:
`ReactionsModel` and `RepostsModel` had 99% the same code, so I refactored it into a common parent class where you can specify the `NostrKind`. I found an unused class called `EventsModel` which seemed like an apt name, so I used that for the parent class. Can rename it if that isn't clear or accurate enough or you don't like it.

Verified that the `RepostsView` and `ReactionsView` still function as expected.


jb55:
this is great, thx!

PR #652, opened 2023-02-19
Fix profile banner blocking touch in horizontal orientation

OlegAba:
The frame of the clipped banner view is incorrect in the horizontal orientation and is blocking touches to views under it. I haven't figured out how to solve the frame size yet but this should be a sufficient fix for now.

Before             |  After
 |

### DIFF
--- a/damus/Views/ProfileView.swift
+++ b/damus/Views/ProfileView.swift
@@ -156,6 +156,7 @@ struct ProfileView: View {
 
         }
         .frame(height: bannerHeight)
+        .allowsHitTesting(false)
     }
     
     var navbarHeight: CGFloat {


jb55:
I can never see your before and afters for some reason

OlegAba:
@jb55

Does it work when not using tables?

### Before
https://user-images.githubusercontent.com/19398259/220167922-7f699dc8-d28c-4fab-b908-2bf08a50785d.mov
### After
https://user-images.githubusercontent.com/19398259/220167945-26448bb6-8ee7-4741-a4dc-a71df76ff4a1.mov

</br>
</br>

How about gifs?
Before             |  After
profile-banner-bug-1 | profile-banner-bug-2

jb55:
ohhh this was annoying. thank you

PR #661, opened 2023-02-20
Fix Post View button

ericholguin:
Just keeps the button in the view and disables it if the text editor is empty

postviewbutton


ericholguin:
This was included in https://github.com/damus-io/damus/pull/853

PR #665, opened 2023-02-21
fix disappearing text on Thread view

bryanmontz:
Fixes an issue where the content text disappears when you long-press to access the context menu.

|Before|After|
|---|---|
|before-fix|after-fix|


bryanmontz:
Improved the solution so that it looks better in light mode:

|Before|After|
|---|---|
|bad|good|

jb55:
thanks!

PR #668, opened 2023-02-21
Long-press on zap button to customize zap amount

boston-wine:
It can be inconvenient to return to profile settings to change zap amount before zapping on a specific note, but having a "default" amount is a useful feature in general.

If there were two options, the default as well as a "haptic" long-press on the zap button to set a custom amount before the Lightning invoice is generated, it would cover both instances.

Note, this is my first PR so feedback and constructive criticism is welcome.


boston-wine:
Noticing it looks like I tried to merge 264 commits here. Definitely didn't intend to do that; just trying to share an idea.

Will focus on learning the platform before adding more here.

boston-wine:
Thanks @alltheseas for linking a comparable, preexisting suggestion. Closing this PR for now.

Issue #493, opened 2023-02-02
Search Notes for Keyword [NIP-50][500,000 sats]


alltheseas:
## User Story ##
As a Damus user looking to search for a particular keyword, I would like the capability to search all notes for that keyword, so that I can filter to notes with the keyword I am searching for.

### Acceptance Criteria ###
1. User can search all notes for a particular keyword
2. Damus returns a reverse chronological order filter view with notes containing that keyword
3. Damus highlights keyword
4. Support option for multiple NIP-50 providers
5. Local search https://github.com/damus-io/damus/issues/1760

See NIP-50: https://github.com/nostr-protocol/nips/blob/master/50.md

mockup:
![global note search](https://user-images.githubusercontent.com/64376233/216435692-c1ef62f2-55d7-41a1-a11d-38738b9775c9.png)

brugeman:
There is a bounty here now: https://nostrbounties.com/b/naddr1qq9rzd3cxgungwfjxcmqygpn2m0xrvukg7f3e69jzs9jh2ur0cypps8029dmayk7qfyqgzutm5psgqqqw4rssvx9yw

alltheseas:
https://damus.io/note1k3g32usmezhkgmjrxv39qu6rt6g6eahwkeywltg8603mnltelzmscz29h9

jb55:
Will be able to do this locally with fulltext search in nostrdb

alltheseas:
See search relay by semisol currently for kind1 only

https://damus.io/note1jr7dgwn98da30wynwkunhy8kgslqey6qknhfm576ppq4ceftpcmqalye7f

> relay.noswhere.com
real time, 👀

jb55:
We can search notes for keywords, lets create more granular tasks for the remaining things (nip-50 remote search, highlighting keywords, etc)

PR #671, opened 2023-02-22
Search by QR code

hazeycode:
Added QR capture functionality, accessible from the SearchHomeView.

- Tap the purple QR button to bring up the capture view (requests camera access)
- When a QR code is captured the view will be dismissed and the search field will simply be populated with the string
- Toggle the device flashlight by tapping the button in the top right of the capture view
- The capture view may be dismissed by swiping it or by tapping the X in the top left of the capture view

IMG_59D6EE0ADF52-1


hazeycode:
> @hazeycode you interested in implementing the keyword search ticket: #493?

Looks good. But I only submitted this PR because I "needed" the feature for myself. Your feature is probably a bit more involved than this one and I don't have time right now but I will keep it in mind.

hazeycode:
I was just using it to scan QR codes I was outputting from another program. But yeah! One person can display their QR code from the menu and the other can scan it using this to link up on nostr from in person.

PR #672, opened 2023-02-22
move misplaced error log statement

bryanmontz:
Before this change:
A log statement of "decode failed for ..." was repeatedly logged to the console, causing me to think there was a parsing problem for events. I started trying to debug that issue.

It turned out that it was just that the log statement was misplaced. It was being shown for every event over 2000 characters, because they were being sent to a background thread for decoding.

After this change:
The log statement is in the correct place, and it only shows in the console when there is actually a failure to decode an event.


bryanmontz:
Superseded by https://github.com/damus-io/damus/pull/682

PR #687, opened 2023-02-24
bugfix: make navbar back button change based on the previous page.

jacany:
When viewing a opening note on the notifications page for example, the back button would say Universe even though it wasn't opened there.


jb55:
thanks!

jb55:
this seems to duplicate code (timelineNavItem) can we de-dupe before we merge this?

jacany:
> this seems to duplicate code (timelineNavItem) can we de-dupe before we merge this?

Fixed.

PR #689, opened 2023-02-24
Fix localization issues, add tests, import translations, and add zh-CN and zh-TW

tyiu:


jb55:
Thanks for adding the test!

PR #691, opened 2023-02-24
rich-tagging

scoder1747:
User can see who they are tagging

https://user-images.githubusercontent.com/120697811/221260997-e9e823e2-bb8d-45e0-97fe-89bcbc91c970.mov


jb55:
code looks pretty good! testing now

PR #702, opened 2023-02-27
Wrap event profile display name

OlegAba:
Long display name and username separated by new line

Before             |  After
IMG_4950 | IMG_4948

Short display name and username stay on the same line

Before             |  After
IMG_4951 | IMG_4952


jb55:
so good! thanks

PR #703, opened 2023-02-28
Recommended Relay View improvements

ericholguin:
This PR adds the pay label to the recommended relay view shown in a user's relay settings as well as another user's relay view

relayimpvmnt


jb55:
thank you!

Issue #666, opened 2023-02-21
Disable all animations toggle [Accessibility][UX][Feature Request]


alltheseas:
## User Story ##
As a Damus user who has an accessibility negative user experience with experiencing animations, I would like a setting that allows me to disable all animations, so that I can improve my Damus user experience. 

### Acceptance Criteria ###
1. Damus user has option to disable (and enable) all animations in the `Settings` Menu

Mockup:
![Disable Animations](https://user-images.githubusercontent.com/64376233/220410744-a697571e-cd65-4bd0-8ba7-4beb8ed6027e.png)

Origin
"_`Just sent a ⚡️Tip of 1,000,00 Sats to #[0] for raising a topic to #NostrDevs that will allow his friend who lives with epilepsy to join Nostr. Topic raised on 02/19: Client Accessibility. #NostrDevs: Can a stop all animation feature be created and added to Client Settings features?`_"

![image](https://user-images.githubusercontent.com/64376233/220408603-16e3dce2-15b2-4f30-8b5b-170ca8d285f3.png)

-Gabe & Angela

tyiu:
Good idea. Should be easy to implement.

joelklabo:
Anyone working on this one? I guess that would mean not animating GIFs as well?

tyiu:
> Anyone working on this one? I guess that would mean not animating GIFs as well?

I believe so.

OlegAba:
Hi all, I have a PR up proposing to switch image libraries. I will implement this with whichever library is decided on.

bryanmontz:
Instead of having a separate setting inside Damus for disabling animations, it should simply respect the user's `UIAccessibility. isReduceMotionEnabled` value.

bryanmontz:
@alltheseas 
Thanks for checking and continuing the conversation!

To each their own, but here's why I respectfully disagree:
1. A user with the Reduce Motion setting turned on has already declared that they don't want to see unnecessary motion, so there's no further decision that they should need to make by flipping another switch in each app. Apple's way for their platforms is that things should "just work", meaning the user should have to fiddle with things the least they possibly could and still get their desired result.
2. The developer of any given app respecting that setting by default shows as much care about accessibility settings as does having another switch in the app.

I suppose that as long as the in-app setting's default value is to read the accessibility value, and then give the user the option to override that, it would be fine. I just wouldn't want the app to default to showing animations when the accessibility setting was to not show it.

bryanmontz:
[...]

PR #707, opened 2023-02-28
Add option to disable image animation

OlegAba:
Ref: #666 

The option to disable animated images defaults to `UIAccessibility.isReduceMotionEnabled`. 

IMG_4962


jb55:
yay!!

PR #709, opened 2023-02-28
Use DamusPurple instead of .purple

joelklabo:
| Before | After |
| - | - |
| Simulator Screen Shot - iPhone 14 Pro - 2023-02-28 at 12 48 18 | Simulator Screen Shot - iPhone 14 Pro - 2023-02-28 at 12 48 46 |


jb55:
Is it bad I actually like purple more for mentions/links 😬

joelklabo:
> Is it bad I actually like purple more for mentions/links 😬

No, I was actually thinking that but they're half and half now it looks like, they could all be `.purple`. In the quoted note you can see they are different, unless that was intentional

joelklabo:
I'm no purple expert 🤣

PR #722, opened 2023-03-02
Translations

tyiu:


damus/Views/Zaps/CustomizeZapView.swift
```diff
@@ -84,28 +84,27 @@ struct CustomizeZapView: View {
...
-            return "No zaps are sent, only a lightning payment."
+            return NSLocalizedString("No zaps are sent, only a lightning payment.", comment: "Description of non-zap type where sats are sent to the user's wallet as a regular Lightning payment, not as a zap.")
         }
     }
     
     var ZapTypePicker: some View {
         Picker(NSLocalizedString("Zap Type", comment: "Header text to indicate that the picker below it is to choose the type of zap to send."), selection: $zap_type) {
             Text("Public", comment: "Picker option to indicate that a zap should be sent publicly and identify the user as who sent it.").tag(ZapType.pub)
             Text("Private", comment: "Picker option to indicate that a zap should be sent privately and not identify the user to the public.").tag(ZapType.priv)
-            Text("Anon", comment: "Picker option to indicate that a zap should be sent anonymously and not identify the user as who sent it.").tag(ZapType.anon)
```
tyiu, line 105: @jb55 I understand why you chose to truncate `Anonymous` (because otherwise the text in the segmented picker doesn't fit on the screen), but this doesn't translate well and isn't obvious to some what it means. Also, for languages that translate to text that is longer than English, it gets cut off anyway. I changed it to a menu picker style.

jb55:
thx

PR #729, opened 2023-03-02
Additional Delete Confirmation

ericholguin:
Adds additional confirmation and information when users press delete account. Also adds sign out to config menu seeing as many users expect that to be there.
signout  deleteaccount


jb55:
lol perfect thank you

PR #731, opened 2023-03-04
Double tap custom zap

ericholguin:
This PR just makes zaps work like before where on tap automatically opens the wallet. And if the zap button is double tapped it will open the custom zap menu.

**NOTE:** _I know theres been discussion on a long press, I think theres potential for some fun stuff for that. Where as long as a user is long pressing the zap button it shows a sheet where its incrementing the amount of sats and the phone vibrates the whole time or possibly other ideas._ 

doubletapzap


jb55:
I don't think users would ever know to double-tap to find this menu

jb55:
Went with a long press gesture:

fc83cd4db7c43153b1abcf67231cd0d01b9a43c5

PR #732, opened 2023-03-04
Thread Caching

jb55:
Make threads load instantly and use cache when building parent and child events

Changelog-Added: Threads now load instantly and are cached

Last things todo:

- [ ] Ensure we re-subscribe when switching selected events to load zaps, etc
- [x] Scroll to selected event


jb55:
merging this even though it's not perfect. it too good not to.

PR #736, opened 2023-03-05
prune EventCache when iOS fires memory warning

bryanmontz:
The new `EventCache` is fantastic for speed in Damus. However it can very quickly cache hundreds or thousands of events in memory. When Damus is backgrounded by the user, iOS may fire memory warning notifications to reclaim memory. If Damus does not return enough memory to the OS, it will kill the app. However if Damus does return enough memory to the OS, it will get to keep running. This will make Damus a better OS citizen, and it will be more reliable for the user and resume rather than launching from scratch more often.

You can simulate a memory to see it work:
Debug_and_Menubar_and_iPhone_14_Pro

### DIFF
--- a/damus/Util/EventCache.swift
+++ b/damus/Util/EventCache.swift
@@ -5,17 +5,23 @@
 //  Created by William Casarin on 2023-02-21.
 //
 
+import Combine
 import Foundation
+import UIKit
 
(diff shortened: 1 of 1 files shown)


jb55:
thanks. seems reasonable

Issue #700, opened 2023-02-27
Repost Reply does nothing


MonkeyBars3k:
iOS app

Repost triggers popup on the Posts timeline but hitting the Repost button on a Reply in the Posts & Replies timeline just shows the tap animation and never launches the popup.

MonkeyBars3k:
Video linked shows user clicking into thread _then_ attempting Repost.

Bug only happens when tapping Repost button directly _on a Reply in Posts & Replies timeline_.

Just tested again and bug is still present

MonkeyBars3k:
Hm, is that video of the Posts & Replies timeline of someone's profile? I'm referring to the main timeline...

App Store Damus

MonkeyBars3k:
Fixed

PR #738, opened 2023-03-06
Fix repost alert

OlegAba:
Issue ref: #700 

This PR will fix the repost button randomly not showing an alert confirmation. 

Alerts should exist in the same hierarchy to work properly. In our case, the `EventActionBar` alert was being blocked by `ContentView` alerts.


jb55:
omg thx

Issue #728, opened 2023-03-02
Cursors jumps when writing a note when pressing "Return" [Bug][UX]


lima46zulu:
When clicking on "return" the cursor jumps to the end of the text. See video. 

This also happens when selecting say the last sentence of a paragraph and hitting enter. The text will be removed, as expected, but then after the cursor will jump to the end of the text. 

https://user-images.githubusercontent.com/109115349/222574168-84d762f6-4c57-48ca-986b-047b157e940b.mp4

dmnyc:
Confirmed - same issue in replies.
https://github.com/damus-io/damus/issues/747

PR #742, opened 2023-03-06
Fix Issues #728,747 - cursor jumping after pressing Return

gladiusKatana:
Fix bug reported in Issue #728 & Issue #747

Bug: 
-On hitting Return in a post, between 2 existing lines, cursor was jumping to bottom of screen

Fix:
-In `TextViewWrapper` / `updateUIView`, temporarily disable scrolling, set the `selectedRange`, then re-enable scrolling
-Adjust cursor appropriately after user-tagging actions: pass the character-count change (ie. full tagged username minus that of the substring that was typed), into `updateUIView`

Other cursor UX enhancements proposal...
-More fine-tuning of cursor position depending on latest action (eg paste, return to a draft, etc), could be easily handled in other PR(s) using the small ObservableObject class added here

Demo

https://user-images.githubusercontent.com/47217795/236730510-29a72e57-3158-42b6-b72e-b2cbeba189c5.mov


gladiusKatana:
@jb55 
quick question too, are there any threads elsewhere for technical discussion? eg. telegram/slack etc.  (i'm pretty new here both as contributor & a user)

jb55:
thanks!

yeah you can use the damus telegram (https://t.me/damusios) or just coordinate on nostr with the devs

gladiusKatana:
> reverting this as well since it introduces some other bugs :( it doesn't seem to jump to the end when you mention someone anymore

Ok updated to put cursor in correct position after a @mention. Pinged you on nostr. 
Did not see any _other_ side effects in testing, but LMK if any were missed; i expect they should be pretty simple to fix (as should a couple other small cursor UX enhancements i noticed that are outside the scope of this ticket)

dmnyc:
Is this fix ready to deploy?

jb55:
Oops this was merged? Not sure why this is still open.

PR #752, opened 2023-03-07
Alert for invalid bitcoin tip address

scoder1747:
**Alert for a tip address that is non-email like or not starting with LNURL**
Simulator Screen Shot - iPhone 14 - 2023-03-07 at 17 38 36


benthecarman:
would be better if it tried to generate an invoice from it, that way it could tell if it will actually work

PR #753, opened 2023-03-07
Fix selected event text padding

OlegAba:
Before             |  After
IMG_4996 | IMG_4997


jb55:
subtle! thanks

PR #755, opened 2023-03-08
Minor text changes

ericholguin:
This PR adds some minor text changes to events:

- Adjusted repost font size and weight to use a heavier font and size that matches profile name
- Adjusted ellipsis color to use light gray in light mode and medium gray in dark mode
- Added dot operator to separate event time from profile name
- Use the same color form profile name as event time

**|--------------------Before-----------------------------|--------------------------After-------------------------|**

 

**|--------------------Before-----------------------------|--------------------------After-------------------------|**


jb55:
looks great!

PR #758, opened 2023-03-08
Fix KF options order

OlegAba:
I ordered the options incorrectly when I last refactored this code. According to the docs, backgroundDecode should be set after the processor.


jb55:
hmmm ok thanks!

PR #759, opened 2023-03-08
ContentView.swift: fix nav title text presentation

RandyMcMillan:
When the damus-home image is presented - the .frame is set to 30x30
We set the .frame to 300x30 to handle other cases.


RandyMcMillan:
Screen Shot 2023-03-08 at 5 21 02 PM

RandyMcMillan:
image

RandyMcMillan:
am i the only person that sees this?

Issue #725, opened 2023-03-02
Respect "Dynamic Type" (Default font size) for Compose view


lima46zulu:
The text in the compose window (new or replies) is one of the only text element in the app that doesn't respect the dynamic type setting. Currently the text in the compose window is too small. 

- Composing new notes
- Writing a reply

bryanmontz:
Fixed in PR https://github.com/damus-io/damus/pull/761

bryanmontz:
@alltheseas This Issue can be closed now because the PR was merged.

bryanmontz:
You're right. It makes sense to wait until you can test it in a build. Thanks for all your help!

alltheseas:
Closing as part of a backlog triage — we're narrowing focus to the most actionable issues to reduce noise for contributors.

If this is still relevant, please reopen with updated context or reproduction steps. Thanks for reporting!

PR #761, opened 2023-03-09
fix: unexpected font size on PostView

bryanmontz:
Fixes an issue where the text field the user types in for posts and replies does not respect the user's Dynamic Type setting (selected font size).

|Before|After|
|---|---|
|bad-1|good-1|
|bad-2|good-2|

Fixes issue: https://github.com/damus-io/damus/issues/725


jb55:
thanks!

PR #762, opened 2023-03-09
run-tests.yaml: trigger push/pull all branches

RandyMcMillan:
run-tests.yaml: trigger push/pull all branches


RandyMcMillan:
This allows dev to use the CI on their repo topic branches.

joelklabo:
makes sense 👍

PR #763, opened 2023-03-09
Fix keyboard related bugs

OlegAba:
Sending a wallet link from the custom zap view while the keyboard is up breaks keyboard avoidance. The solution is to dismiss the keyboard before sending the app link.

Before             |  After
keyboard-1 | keyboard-2

</br>

Also removed keyboard avoidance for the tabbar

Before             |  After
tabbar-1 | tabbar-2

### DIFF
--- a/damus/ContentView.swift
+++ b/damus/ContentView.swift
@@ -214,7 +214,6 @@ struct ContentView: View {
                 }
             }
         }
-        .ignoresSafeArea(.keyboard)
     }
     
     var MaybeSearchView: some View {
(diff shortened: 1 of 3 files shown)


damus/ContentView.swift
```diff
@@ -257,49 +256,47 @@ struct ContentView: View {
         VStack(alignment: .leading, spacing: 0) {
             if let damus = self.damus_state {
                 NavigationView {
-                    ZStack {
```
OlegAba, line 260: Removed unnecessary ZStack (260 - 301)

Issue #741, opened 2023-03-06
Space after #hashtag in search displays no results [bug][UX]


alltheseas:
Thanks for fixing [hashtag search case sensitivity ](https://github.com/damus-io/damus/issues/708)on #dAmUs @RandyMcMillan​ 🙏  

Works for me now.

I did discover a small bug: if I add a space after hashtag text, nothing appears. 

I suppose the expected behavior should be disregard the extra space, and show results of hashtag string without the extra space?

PR #773, opened 2023-03-12
Fix Issue #741 - Space after #hashtag in search displays no results

gladiusKatana:
Fix Issue #741 

Fix:
-Filter out whitespaces from the `Search.hashtag` value obtained from the search query (ie `search`, the source-of-truth)
-As expected, the proper hashtag shows below in the NavigationLink - with no spaces

Screenshots:
[Search entry]
Screenshot 2023-03-12 at 6 00 56 PM

[On tap of link]
Screenshot 2023-03-12 at 6 27 11 PM


jb55:
thanks!

Issue #769, opened 2023-03-10
Mentions don't work in the middle of a note when writing a new note 


lima46zulu:
When adding a mention in the middle of the note, adding the @ won't trigger the username list to appear and the @username will just be text. 

For example, if I type: 

> Hello
> How are you fren?

Then adding a mention after Hello will not work. The workaround is to delete all the text after the mention and rewrite or past it. Mentions only seem to work when the @ is the last character of a note. 

![IMG_5137](https://user-images.githubusercontent.com/109115349/224406837-22a57630-aba1-4152-8936-84a6acf27fef.jpeg)

dmnyc:
I recorded an example of the issue I mentioned above.
https://nostr.build/av/nostr.build_b7306f9831d2db9e1237ff9a2fa386374be36ed475bb785277f88e699a775866.mov

alltheseas:
https://github.com/damus-io/damus/pull/775

alltheseas:
This seems resolved in 1.6 (19)

PR #775, opened 2023-03-13
Fix Issues #769 & #728 : "Mentions don't work in the middle of a note [...]" & "cursor jumps to the end of the text."

gladiusKatana:
_-Fix issue #728 : "When clicking on "return" the cursor jumps to the end of the text."
-Fix issue #769 : "When adding a mention in the middle of the note, adding the @ won't trigger the username list to appear and the @username will just be text."_

-In `get_searching_string(...)`, return first word prepended with "@" that is not _also_ a previously-selected @mention
-In `UserSearch` / `.onTapGesture`, delete the above returned search-word & insert in its place the full, formatted @mention
-Fix cursor position: In `TextViewWrapper` / `updateUIView(...)`, update how/when `uiView`'s properties are set eg. `uiView.selectedRange` is set conditionally on the action most recently taken (eg. cursor to right of a selected tag)

*
* Handling spaces in usernames:
-To make the UX of the dynamic search smooth even for usernames containing spaces (" ") ... 
In a username's 'tag attributed-string' inserted upon selection (not the username itself), replace all " " occurrences w/ Unicode character `0x2004`, which looks identical in the UI. (Note, of course username links to profiles still work in a completed post as expected; and other alternate username characters (eg. spaces 0x00A0, 0x2002, 0x2003) do not cause the above spaces issue

Demo

https://user-images.githubusercontent.com/47217795/234422379-2ddde56e-83a8-4a82-b76a-c0086a826fe4.mov


jb55:
awesome. thank you!

jb55:
had to revert this since the tag popup doesn't go away after completing a mention

gladiusKatana:
@jb55 Ok fixed the tag popup / dismissal. (Updated video)
BTW this logic should lend itself to supporting tagging multiple users, after some minimal refactoring :)

gladiusKatana:
Ok i merged master in -- still works.
Note, there's also a side benefit if we enable tagging multiple users (commit 978dec6... reverted for now),
this 2-line commit will prevent the new `EmptyUserSearchView` presentation upon tagging a user in the middle of a note 
(see video ie this frame):

image

jb55:
is this ready for another review ?

gladiusKatana:
@jb55 yeah, I've tested

> is this ready for another review ?

Yeah! I'll update the video right now and ping you on Damus (saw your post on that - makes sense).

gladiusKatana:
> is this ready for another review ?

Yeah :) Already pinged you on Damus as you know ✌🏾

jb55:
thank you for pinging me on nostr, otherwise I would never see this. looks pretty good now. testing

jb55:
this is getting complicated and messy. surely there must be a better way to do this...

jb55:
merged, thanks!

PR #776, opened 2023-03-14
improvement-zap-vibration

scoder1747:
**This change is about making the zap-vibration more impactful.**


jb55:
now that we have notifications I might just remove this vibration hack

PR #777, opened 2023-03-14
fix "Replying to..." issues and improve related tests

bryanmontz:
This PR fixes two issues with the "Replying to..." string.
1. Zero-length usernames are rendered in the string.
2. Duplicate names. 

I also added a test for the duplicate names one and refactored the duplicated code in the tests.

|Before|After|
|---|---|
|before|after|


jb55:
great!

PR #779, opened 2023-03-15
configurable-custom-zap-sheet

scoder1747:
**Story:**
The new users might not know anything about Zap. Considering the new users, it is always good to show the users the custom zap sheet by default. Once the users are habituated, they can turn off the sheet by switching off the toggle (located by the end of the sheet) and immediately defining their default zap amount on the same zap sheet (next to the toggle). That being said, the long pressing of zap icon will always trigger the custom zap sheet at given point of time. 

**Relevant screenshots (two):**
Simulator Screen Shot - iPhone 14 - 2023-03-14 at 20 04 07
Simulator Screen Shot - iPhone 14 - 2023-03-14 at 20 04 18

### DIFF
--- a/damus/Components/ZapButton.swift
+++ b/damus/Components/ZapButton.swift
@@ -75,19 +75,22 @@ struct ZapButton: View {
                     .font(.footnote.weight(.medium))
             })
             .simultaneousGesture(LongPressGesture().onEnded {_  in
-                guard !zapping else {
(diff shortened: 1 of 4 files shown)


jb55:
yeah I think this is fine, do we need the extra settings at the bottom? Perhaps we can just keep things in settings and direct them there? Seems weird to have to places to configure this.

PR #780, opened 2023-03-15
Move note content rendering to async task

OlegAba:
A minor optimization to `NoteContentView` init. Test for loading `1000` rows:

#### Before
`avg time 0.19702434539794922 milliseconds`

#### After
`avg time 0.00253140926361084 milliseconds`


jb55:
I think I moved this to the constructor so that artifacts wouldn't pop-in

OlegAba:
@jb55, I disagree with this decision. I think the visual change of artifacts loading is preferable to the entire app hanging. The change is only noticeable when the loading is expensive (many artifacts), in which case the alternative results in a significant hang. If we want to further optimize the app and resolve the remaining hangs/hitches, we should not have expensive operations in the view init and body.

jb55:
I agree but the popping is insanely annoying. ideally the artifacts would be ready somehow before the view is constructed?

OlegAba:
@jb55 ok will look into an alternative.

PR #784, opened 2023-03-16
fixes issue where pasteboard.co image links don't render

bryanmontz:
This PR fixes an issue where images from pasteboard.co do not render in `EventView`. It appears to be because although the URLs from pasteboard end with common image format file extensions, the content is actually html rather than image data. If we instead treat it as a non-image link, the meta information from pasteboard.co will be rendered in the view.

|Before|After|
|---|---|
|bad|good|

### DIFF
--- a/damus/Views/NoteContentView.swift
+++ b/damus/Views/NoteContentView.swift
@@ -239,9 +239,20 @@ func render_blocks(blocks: [Block], profiles: Profiles, privkey: String?) -> Not
 }
 
 func is_image_url(_ url: URL) -> Bool {
-    let str = url.lastPathComponent.lowercased()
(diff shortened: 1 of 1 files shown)


damus/Views/NoteContentView.swift
```diff
@@ -239,9 +239,20 @@ func render_blocks(blocks: [Block], profiles: Profiles, privkey: String?) -> Not
...
 
 func is_image_url(_ url: URL) -> Bool {
-    let str = url.lastPathComponent.lowercased()
-    let isUrl = str.hasSuffix(".png") || str.hasSuffix(".jpg") || str.hasSuffix(".jpeg") || str.hasSuffix(".gif")
-    return isUrl
+    // Some media hosts provide links to an "image preview" with
+    // ads and other content instead of directly to an image file,
+    // which causes problems with rendering in an image view.
+    let hostsThatWrapImages = [
+        "pasteboard.co"
```
jb55, line 246: this seems like an impossible thing to maintain

PR #792, opened 2023-03-19
fixes issue where image links don't render

bryanmontz:
This fixes the issue where pasteboard.co image links (or any host) don't load at all, but in a better way that does not require hard-coding known image host sites that wrap images with ads or other content.

The basic idea is:
If a link ends with the image file endings (.jpg, .gif, etc), then we'll first try to load it into a `KFAnimatedImage` as before. Thankfully it has an `onFailure` modifier that we can listen for. If it fails, we will add the url to a cache of failed urls and then reload it as a non-image url, which generally brings up the image from the site's metadata.

|Before|After|
|---|---|
|bad-profile|good-profile|
|bad-bookmarks|good1|
|bad-dms|good-dms|


jb55:
looks good, we probably could have just thrown this in the Previews cache or something instead of adding another thing to DamusState. not going to worry about it though and can fix it after

bryanmontz:
Sure, easy enough to move to wherever you think makes the most sense.

PR #796, opened 2023-03-21
typo-image-uploader-name

scoder1747:
**Resolved complaints on image uploader naming. They have confirmed that this change looks good in the noteid: note1g2c00jnvcfl746ztkpqp6j5de7mu7w56zy47er7u7lspn7zk3clsd9ccml** 

Simulator Screen Shot - iPhone 14 Plus - 2023-03-20 at 22 40 08


damus/Views/AttachMediaUtility.swift
```diff
@@ -140,24 +140,24 @@ extension NSMutableData {
...
-        case .nostrBuild:
+        case .nostrbuild:
             return "\"fileToUpload\""
-        case .nostrImg:
+        case .nostrimg:
             return "\"image\""
         }
     }
 
     var displayImageUploaderName: String {
```
tyiu, line 155: It doesn't seem like you're using this variable. Remove?

damus/Views/AttachMediaUtility.swift
```diff
@@ -170,26 +170,26 @@ enum ImageUploader: String, CaseIterable, Identifiable {
 
     var model: Model {
         switch self {
-        case .nostrBuild:
-            return .init(index: -1, tag: "nostrBuild", displayName: NSLocalizedString("NostrBuild", comment: "Dropdown option label for system default for NostrBuild image uploader."))
-        case .nostrImg:
-            return .init(index: 0, tag: "nostrImg", displayName: NSLocalizedString("NostrImg", comment: "Dropdown option label for system default for NostrImg image uploader."))
+        case .nostrbuild:
+            return .init(index: -1, tag: "nostr.build", displayName: NSLocalizedString("nostr.build", comment: "Dropdown option label for system default for nostr.build image uploader."))
```
tyiu, line 174: You can just remove the `NSLocalizedString` and use the string directly at this point. There's no reason to localize a domain name.

scoder1747: Makes sense. Removed.

damus/Views/AttachMediaUtility.swift
scoder1747: This computed var was used for generating the image uploader's name in the picker. Not sure how it's use case got removed

jb55:
merged a fix for this already, thx tho!

PR #800, opened 2023-03-24
Relay Config View changes

ericholguin:
Several changes to the relay config view and associated views:

- Bigger titles
- Use Bitcoin Logo
- Edit Button to show add/removal buttons
- Add Relays in the same view
-  Connect/Disconnect from relay detail view 
- Connection Icon changes

relay-edit relay-actions
add-new user-relays


ericholguin:
> I'm not against the image but we already have bitcoinsign.circle.fill in sfsymbols. although the svg looks better for sure.
 Up to you, I can switch it
btcsystem

PR #803, opened 2023-03-24
zap-notification

scoder1747:
Zap in-app notification:
Simulator Screen Shot - iPhone 14 Plus - 2023-03-24 at 17 59 09

### DIFF
--- a/damus/ContentView.swift
+++ b/damus/ContentView.swift
@@ -320,6 +320,7 @@ struct ContentView: View {
         .onAppear() {
             self.connect()
             setup_notifications()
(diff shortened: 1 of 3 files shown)


jb55:
trying this now!

damus/ContentView.swift
```diff
@@ -774,8 +775,21 @@ func setup_notifications() {
             return
         }
     }
+
 }
 
+func setup_notifications_permission() {
```
jb55, line 781: not sure what the point of this is, isn't this already handled by `setup_notifications` ?

PR #804, opened 2023-03-25
Fix notifications to show previews and allow tapping reaction text to navigate to threaded event

tyiu:
Simulator Screen Recording - iPhone 14 Pro - 2023-03-24 at 22 08 49


jb55:
thank you!

PR #815, opened 2023-03-27
Implement zapping zap tags

bu5hm4nn:
This branch implements per event "zap" tags as proposed in PR https://github.com/nostr-protocol/nips/pull/402

It also adds a bunch of comments for developers desiring to understand how damus processes kind 9735 events.

### DIFF
--- a/damus/Models/HomeModel.swift
+++ b/damus/Models/HomeModel.swift
@@ -153,33 +153,50 @@ class HomeModel: ObservableObject {
         return
     }
 
+    /// Handle zap notifications (kind 9735)
     func handle_zap_event(_ ev: NostrEvent) {
-        // These are zap notifications
+        /// The p-tag on the 9735 event points to the profile who authored the event that got zapped
         guard let ptag = event_tag(ev, name: "p") else {
             return
         }
         
+        /// our keypair is the keypair of the current damus user
+        /// it will be used to identify zaps on posts of the current user
+        /// so that appropriate notifications (like vibrating) can be dispatched
         let our_keypair = damus_state.keypair
+        
+        /// lookup if we already know the zapper in our local cache
+        /// in this case we don't have call the http endpoint again and can handle it and be done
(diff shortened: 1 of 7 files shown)


bu5hm4nn:
@jb55 let me know when you are ready to consider this and I will fix the conflicts. I just don't have the time to do it over and over again.

jb55:
I think we're only going to implement pubkey zap targets

bu5hm4nn:
I don't even understand what benefit pubkey zap tags give over what we currently have already, but I'm not here to argue, feel free to close this PR if you don't need it.

Zapstr will go online in about a month and that will use LNURL zap tags with replaceable events, but they will be using a new event kind for music, so that would be a whole new question for you anyways if you want to add basic music playback or not (like Telegram App does for example). Maybe depends on user demand as well.

jb55:
anything other that pubkey zaps is a security problem. anyone could say "hey all zaps are forwarding to fancybob13@walletofsatoshi.com" which is owned by Bob. In damus we can't do anything with this or prove to other users that the money is indeed getting forwarded to Bob who owns fancybob13@walletofsatoshi.com. I guess we would have to build a lookup of all the profiles we've seen and hope we have bob's profile so that it matches so that we can display "forwarded to bob" on the ui?

No, ln address zap targets are dumb. Only pubkey zap targets make sense in the context of forwarding zaps in nostr.

jb55:
In the future I would like to see zap splits defined in terms of pubkeys within notes themselves, so that damus can display the splits securely in the UI.

bu5hm4nn:
You are always trusting the note author when you zap them. That does not change when you zap a profile he specifies. The only scenario this makes any difference is if someone claims he is raising money for lets say DerekRoss and you want to verify it really goes to him. But the better solution for that scenario is to just repost DerekRoss own note so the zaps go to him. Or use lightning fundraising tools directly.

PR #830, opened 2023-03-29
removed-zero-width-character

scoder1747:
This removes zero-width-character u{200B} before submitting the post. The reason it is being used is because it allows user to place comma or hyphen immediately next to @mentions without passing the link attributes to these characters while composing their post in post view


jb55:
I don't understand this ...

PR #837, opened 2023-03-30
Add support for nostr: bech32 urls in posts and DMs

bmewj:
I've added the necessary C and Swift code to parse and display `nostr:npub` `nostr:note` `nostr:nevent` `nostr:nprofile` `nostr:naddr` and `nostr:nrelay` correctly in posts and DMs. The hope is to extend support for [NIP-27](https://github.com/nostr-protocol/nips/blob/master/27.md), where `p` and `e` tags are now optional. Users who want to publish notes referencing pubkeys or events in the content without having the references tagged can use the standard `nostr:` syntax instead of `#[idx]`.

This will come in handy later in order to remove `#[idx]` references entirely from DMs to stop mentions from leaking into the public.

### DIFF
--- a/damus-c/bech32_mention.c
+++ b/damus-c/bech32_mention.c
@@ -1,0 +1,156 @@
+//
+//  bech32_mention.c
+//  damus
+//
+//  Created by Bartholomew Joyce on 2023-04-03.
+//
+
+#include "bech32_mention.h"
+#include "bech32.h"
+#include <stdlib.h>
+#include <string.h>
+
+#define TLV_SPECIAL 0
+#define TLV_RELAY 1
+#define TLV_AUTHOR 2
+#define TLV_KIND 3
+
+int bech32_mention_parse(bech32_mention_t *mention, const char* str, int len) {
+
+    char prefix[len];
+    u8 words[len];
+    size_t words_len;
+    size_t max_input_len = len + 2;
+
+    if (bech32_decode(prefix, words, &words_len, str, max_input_len) == BECH32_ENCODING_NONE) {
+        return 0;
+    }
+
+    memset(mention, 0, sizeof(bech32_mention_t));
+    mention->kind = -1;
+    mention->buffer = (u8*)malloc(words_len);
+
+    size_t data_len = 0;
+    if (!bech32_convert_bits(mention->buffer, &data_len, 8, words, words_len, 5, 0)) {
+        goto fail;
+    }
+
+    // Parse type
+    if (strcmp(prefix, "note") == 0) {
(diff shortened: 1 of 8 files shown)


bmewj:
Before
Simulator Screen Shot - iPhone SE (3rd generation) - 2023-03-30 at 02 58 14

After
Simulator Screen Shot - iPhone SE (3rd generation) - 2023-03-30 at 02 57 19

`nostr:naddr` and `nostr:nrelay` are being parsed correctly behind the scenes, but I haven't added any UI for them yet as there doesn't seem to be support for such URLs yet in Damus as of yet.

jb55:
doesn't seem to be parsing in DMs ?

bmewj:
> doesn't seem to be parsing in DMs ?

Hmmm. You sure? Are you including the `nostr:` prefix?

bmewj:
> I think we need tests and tav parsing separated into their own functions

For starters I've moved all the new parsing code out into a separate file.

Not sure what you mean by separating tests from the tlv parsing. By testing you mean the bech32 decode and then the checking of the `prefix`?

jb55:
looking at this again today

jb55:
the main use case for me would be to have these kinds of mentions in DMs yet it doesn't work there.

jb55:
IMG_1159

bmewj:
So strange, I retyped your messages on my side and I got the following:

Simulator Screen Shot - iPhone SE (3rd generation) - 2023-04-07 at 22 21 30

Could it be the iOS version? I'm testing on the iPhone SE (3rd generation) Simulator

bmewj:
Tried also on iPhone 14 Pro Simulator for good measure. Also works. I'll have to sort out my developer account to run it on an actual device, but for now that's the best I can do.

Simulator Screen Shot - iPhone 14 Pro - 2023-04-07 at 22 32 33

jb55:
I'll step through and try to figure out why it's not working

jb55:
nm I just had an extra space 🤦🏻‍♂️

jb55:
thanks I'll merge this soon I'm just refactoring it a bit

jb55:
> > I think we need tests and tav parsing separated into their own functions
> 
> For starters I've moved all the new parsing code out into a separate file.
> 
> Not sure what you mean by separating tests from the tlv parsing. By testing you mean the bech32 decode and then the checking of the `prefix`?

fwiw this is what I meant:

https://github.com/damus-io/damus/blob/13354b0eb5e88a03bb9371b8c79dbfef6b0866e0/damus-c/nostr_bech32.c

PR #843, opened 2023-03-30
Refactor auto-translations and add caching

tyiu:
- All translations are cached in memory
- If auto-translate is turned off but a translation service is enabled, `Translate Note` is shown (even when note is not selected), and will only make a network request when tapped

### DIFF
--- a/damus/Util/Translator.swift
+++ b/damus/Util/Translator.swift
@@ -20,18 +20,28 @@ public struct Translator {
         self.userSettingsStore = userSettingsStore
     }
 
-    public func translate(_ text: String, from sourceLanguage: String, to targetLanguage: String) async throws -> String? {
+    /**
+     Translates a string from source language to target language.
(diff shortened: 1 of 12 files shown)


damus/Util/Translator.swift
```diff
@@ -68,10 +70,9 @@ public struct Translator {
 
         struct RequestBody: Encodable {
             let text: [String]
-            let source_lang: String
             let target_lang: String
         }
-        let body = RequestBody(text: [text], source_lang: sourceLanguage.uppercased(), target_lang: targetLanguage.uppercased())
```
tyiu, line 74: This is a notable change in behaviour. Previously, we were relying on Apple's predicted language as the source language. We're now just letting DeepL do it on their side as it's integrated within the translate API, and I'm hoping that DeepL's online prediction is better than Apple's offline one.

PR #844, opened 2023-03-30
Replace deprecated UNNotificationPresentationOptions.alert with .banner and .list

tyiu:


jb55:
ah I tried to do this and it didn't work, I only did .banner which is probably why.

Issue #851, opened 2023-04-01
blank nip-05 error message in profile edit view [bug]


alltheseas:
_what happens_

In profile edit view, I have a blank NIP-05 field, and I receive an error message below the NIP-05 field

> "" is an invalid NIP-05 identifier. 

_what I think should happen_

In profile edit view, if I have a blank NIP-05 field,there should be no error message displayed. 

![image](https://user-images.githubusercontent.com/64376233/229267317-8d61f7bb-c91e-44a9-a151-bafc1e77e673.jpeg)

bryanmontz:
Fixed in PR https://github.com/damus-io/damus/pull/852

PR #852, opened 2023-04-01
don't show NIP-05 validation error message when field is empty

bryanmontz:
Fixes https://github.com/damus-io/damus/issues/851 where the NIP-05 validation error is shown even when the field is empty.

|Before|After|
|---|---|
|broken|fixed|


jb55:
thanks!

PR #854, opened 2023-04-02
sort bookmarks by event created date descending

bryanmontz:


jb55:
I think we should just have them in the order that they are added. right now it adds to the bottom which is not good, we just need to prepend them.

bryanmontz:
Yeah, good plan. I'll do that.

bryanmontz:
updated

jb55:
perfect. thank you!

PR #857, opened 2023-04-03
Nostrcheckme

quentintaranpino:


quentintaranpino:
I'm sure it doesn't work, I just added nostrcheck to all the methods, but I need to collect the data returned by the server (in the form of JSON) and then get the URL, the result, etc. from there.

Is there someone that can help me with that?

quentintaranpino:
Our API: https://github.com/quentintaranpino/nostrcheck-api

jb55:
> I'm sure it doesn't work, I just added nostrcheck to all the methods, but I need to collect the data returned by the server (in the form of JSON) and then get the URL, the result, etc. from there.
> 
> Is there someone that can help me with that?

can you make this a draft PR if it's not done?

quentintaranpino:
> > I'm sure it doesn't work, I just added nostrcheck to all the methods, but I need to collect the data returned by the server (in the form of JSON) and then get the URL, the result, etc. from there.
> > Is there someone that can help me with that?
> 
> can you make this a draft PR if it's not done?

Of course. 😀

jacany:
apikey needs to be included too, why inside multipart form data and not the authorization header?

quentintaranpino:
> 

It is for legacy compatibility, in the new version the API key is no longer needed and it is authenticated via NIP98 in the header. 

If v0 of the API is implemented it is necessary that the apikey is in the body.

PR #860, opened 2023-04-03
Empty View when tagged user isnt found

ericholguin:
Simulator Screen Recording - iPhone 14 Pro Max - 2023-04-03 at 17 09 17


jb55:
nice, thx. 1701dbdfb94935fd7c0a1e8d4f3255e345d9ebb2

PR #880, opened 2023-04-05
deep-link-for-local-notifications

scoder1747:
Deep Linking to Local Notifications (Mentions, Likes, Reposts, DMs, Zaps).

https://user-images.githubusercontent.com/120697811/230225286-f37a5e24-08ba-432d-94ce-535cb78d1d5a.mov


damus/ContentView.swift
```diff
@@ -156,6 +158,13 @@ struct ContentView: View {
                     EmptyView()
                 }
             }
+            // Navigation when clicking local notification
+            if nav_local_notif_active {
+                let thread = ThreadModel(event: ev_local_notif!, damus_state: damus_state!)
```
jb55, line 163: please don't use !

damus/ContentView.swift
```diff
@@ -156,6 +158,13 @@ struct ContentView: View {
                     EmptyView()
                 }
             }
+            // Navigation when clicking local notification
+            if nav_local_notif_active {
```
jb55, line 162: I don't think you need this if statement

jb55: maybe it should just be 

```swift
if let ev_local_notif {
}
```

damus/ContentView.swift
```diff
@@ -468,6 +477,16 @@ struct ContentView: View {
...
         }
+        .onReceive(handle_notify(.local_notification)) { notif in
+            if let obj = notif.object as? String {
+                if obj == "dm_local_notification" {
+                    selected_timeline = .dms
+                } else if let notif = damus_state?.events.lookup(obj) {
+                    ev_local_notif = notif
+                    nav_local_notif_active = true
+                }
+            }
```
jb55, line 488: prefer this style:

```suggestion
            guard let obj = notif.object as? String else {
                return
            }

            if obj == "dm_local_notification" {
                selected_timeline = .dms
            } else if let notif = damus_state?.events.lookup(obj) {
                ev_local_notif = notif
                nav_local_notif_active = true
            }
```

scoder1747: a good call out!

damus/Models/HomeModel.swift
```diff
@@ -548,7 +548,7 @@ class HomeModel: ObservableObject {
                 self.new_events = notifs
                 if  damus_state.settings.dm_notification,
                     let displayName = damus_state.profiles.lookup(id: self.incoming_dms.last!.pubkey)?.display_name {
```
jb55, line 550: just noticed a ! here, let's fix that while we're here

```suggestion
              if  damus_state.settings.dm_notification,
                     let last = self.incoming_dms.last,
                     let displayName = damus_state.profiles.lookup(id: last.pubkey)?.display_name {
```

jb55: I also thing this logic is wrong since I keep getting the wrong display name, but we can fix that after

scoder1747: addressed, yeah I will also look into it in depth, thx

jb55:
trying this again!

jb55:
I changed the code a bit! it's mostly the same, just refactored a bunch of stuff

PR #893, opened 2023-04-08
Add thread muting

tyiu:
Changelog-Added: Add thread muting


jb55:
I don't think the filtered_zaps logic is correct. I'm seeing 0 amounts in my notification feed for all my zaps now. are you trying to handle the unmute case? I don't think we should worry about that. lets not have all that complicated update logic in notifications... we can stick to filter.

tyiu:
> I don't think the filtered_zaps logic is correct. I'm seeing 0 amounts in my notification feed for all my zaps now. are you trying to handle the unmute case? I don't think we should worry about that. lets not have all that complicated update logic in notifications... we can stick to filter.

I can look into the bug and yes, trying to handle the unmute case. But this is actually a bug even without my changes. I currently see the dot underneath the notifications tab but tapping on it shows nothing new. I believe it's because the notifications data doesn't update properly because relying on counters itself isn't accurate.

jb55:
testing again!

jb55:
this broke DMs... no idea why.

tyiu:
> this broke DMs... no idea why.

Shoot, not sure why either. I can investigate unless you get to it before I do.

jb55:
found the bug. fixing.

jb55:
fixed 042e02d2e447c2edcca08559a90196becb482b40

jb55:
This introduces another bug where it fails to build the notification view

tyiu:
> This introduces another bug where it fails to build the notification view

Damn. How to repro?

jb55:
Fixed in ebba9d3004f7839aeb377c4ba252cecf969c412d

PR #901, opened 2023-04-09
[UX/UI] Following List in Last Added Order

LouisSaberhagen:
Currently the following list is presented in Earliest order ( the contact array is just an append structure where the newest contact is added at the end ).

In my opinion this may be a UX issue because users coming from Twitter expect the opposite behaviour. TWT shows Newest to Earliest user. The typical user stories for this are:

Story A
--
 * You just added someone recently,
 * You don't remember his handle so you quickly jump into the list expecting it to see it first 

Story B
--
 - You added someone
 - You changed views and realized you don't want to follow that person anymore
 - You quickly jump into the list expecting it to see it first and unfollow there

Changes
---
This is a very small change and it doesn't have any impact on performance or the structure of the data. 
It uses the .reversed() method just before presenting the list. This method returns an reverse iterator without allocating any new space or doing any kind of mutation on the data.

damus-screenshot

### DIFF
--- a/damus/Views/FollowingView.swift
+++ b/damus/Views/FollowingView.swift
@@ -57,7 +57,7 @@ struct FollowingView: View {
     var body: some View {
         ScrollView {
             LazyVStack(alignment: .leading) {
-                ForEach(following.contacts, id: \.self) { pk in
+                ForEach(following.contacts.reversed(), id: \.self) { pk in
                     FollowUserView(target: .pubkey(pk), damus_state: damus_state)
                 }
             }


jb55:
this has already been fixed in 1b6224e665bd67193f5d273996d6b859c7ddf19d

LouisSaberhagen:
This PR is about **contacts** not **bookmarks** though 🤔

LouisSaberhagen:
Bumping this. Fix is not about bookmarks but about contacts. @jb55

jb55:
Sorry about that!

LouisSaberhagen:
Failed tests unrelated to this PR. Just checked the Github Action Logs.

jb55:
thanks! ab853c406c1b6c7d71d4ee4bdfa90969213a8582

PR #908, opened 2023-04-11
Improved parsing of bech32 entities

bmewj:
Replaced `consume_until_whitespace()` from the original implementation with `consume_until_non_bech32_character()` to allow for punctuation marks or any other non-bech32 characters to follow a `nostr:` entity without upsetting the parser.

Now things like `Hey, nostr:npub1xtscya34g58tk0z605fvr788k263gsu6cy9x0mhnm87echrgufzsevkk5s!` are parsed correctly.

### DIFF
--- a/damus-c/cursor.h
+++ b/damus-c/cursor.h
@@ -10,6 +10,7 @@
 
 #include <ctype.h>
 #include <string.h>
+#include "bech32.h"
 
 typedef unsigned char u8;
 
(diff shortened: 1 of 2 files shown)


jb55:
thanks! ea14099b627f55f46928d94359202ecbea9cbcae added a test for this: 1f648057d5da6c3c6f06f955f88516ebbe4b800a

PR #911, opened 2023-04-12
Cleanup: nostr bech32 entity parsing

bmewj:
Removed superfluous include

### DIFF
--- a/damus-c/cursor.h
+++ b/damus-c/cursor.h
@@ -10,7 +10,6 @@
 
 #include <ctype.h>
 #include <string.h>
-#include "bech32.h"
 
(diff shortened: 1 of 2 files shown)


jb55:
I guess this would match non-bech32 characters because bech32 chars are not the full set

bmewj:
Yeah, that's what I thought originally, so I checked for only bech32 characters, but then it couldn't parse any of the entities because the prefixes `npub` `note` `nprofile` etc... contain non-bech32 characters. So best you can do is just check for alphanumeric.

jb55:
ok that's fine then

PR #914, opened 2023-04-12
[FIX][UI] Undismissable Keyboard in CustomizeZapView

LouisSaberhagen:
Dismissing the keyboard on CustomizeZapView was not as easy. 

It required tapping a very narrow label, or confunsingly, having to switch to the other textfield or selecting one of the default zaps.

This fix allows the user to dismiss the keyboard by tapping in any area ( specially important given that default behaviour is tapping an  empty area to dismiss ).

Before/After
---
undismissable_keyboardgif

### DIFF
--- a/damus/Views/Zaps/CustomizeZapView.swift
+++ b/damus/Views/Zaps/CustomizeZapView.swift
@@ -200,7 +200,6 @@ struct CustomizeZapView: View {
                     Text("Comment", comment: "Header text to indicate that the text field below it is a comment that will be used to send as part of a zap to the user.")
                 })
             }
-            .dismissKeyboardOnTap()
                 
             Section(content: {
                 ZapTypePicker
@@ -228,6 +227,7 @@ struct CustomizeZapView: View {
             }
         
         }
+        .dismissKeyboardOnTap()
     }
     
     var MainContent: some View {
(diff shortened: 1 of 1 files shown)


alltheseas:
What's your npub @LouisSaberhagen ?

LouisSaberhagen:
> What's your npub @LouisSaberhagen ?

[npub1npd8c6cww4gg446vgygt9efdlwnvufsx8kqtegsc2e9apqa89wvsklljja](https://snort.social/p/npub1npd8c6cww4gg446vgygt9efdlwnvufsx8kqtegsc2e9apqa89wvsklljja)

jb55:
if I remember correctly having the dismiss here causes other issues... will test

jb55:
yup, you can't click the zap button with this change. I remember trying this before.

LouisSaberhagen:
I see. I was testing on an iphone SE and since the screen is smaller I wasn't even seeing the Zap button.

I'll fix it and resubmit the changes.

LouisSaberhagen:
I fixed it. @jb55  

The user can now dismiss the keyboard without interfering with the Zap Button. The solution was to create an overlay that only goes up to ( not including ) the Zap Button. It works on all devices.

taparea

jb55:
Very nice!

LouisSaberhagen:
Thank you

PR #917, opened 2023-04-13
replace Starscream with URLSessionWebSocketTask

bryanmontz:
Removes Damus' dependency on the Starscream library, replacing it with a simple socket implementation using URLSessionWebSocketTask.


jb55:
looks like this fails to reconnect after resume

bryanmontz:
Okay thanks, I'll take a look at that.

bryanmontz:
Okay, try again. Sorry about that. Needed to reset the URLSessionWebSocketTask after any time it gets canceled.

jb55:
cool trying again

jb55:
at some point I get disconnected from everything and my logs get spammed with these messages:

Screenshot 2023-04-15 at 09 54 01

bryanmontz:
Was able to reproduce the issue you saw. With new changes I don't see that anymore. Of course it is expected upon resuming to receive several network errors, but you should no longer be endlessly spammed with them, as the sockets will reset and reconnect.

I noticed that the `RelayConfigView` does not update correctly if you leave the app with it open. When you return to the app, all the relays will show the spinner and never resolve. But if you then back out of the view and go right back into it, it will show the correct state. I'll look into that soon.

bryanmontz:
Oh also I have this code running on my phone today, and I'll use it throughout the day to make sure we reconnect even after longer periods of the app being in the background.

bryanmontz:
Dang, not reconnecting after a longer time. I'll see if I can find what that could be.

bryanmontz:
Okay, I think we're good now. I added a network monitor to RelayPool so that it would attempt to reconnect all relays whenever the network comes back. Now you can launch the app while in Airplane Mode, then turn off Airplane Mode while it is running, and it will quickly make connections and display content. I also tested turning Airplane Mode on and off and on and off while the app is running, and each time it would work as expected, reconnecting relays after re-establishing network connection.

bryanmontz:
I've been running this code on my phone for 24 hours, launching it every now and then to make sure it reconnects and functions properly. Then this morning I did some stress-testing where I tried every combination and order of turning Airplane Mode on and off, Wi-Fi on and off, and LTE on and off, and it always reconnected and worked as expected.

bryanmontz:
@jb55 Merged with master. I've been using this code on my phone all week and it has not failed to connect/reconnect one time. I think this is ready. Again sorry for the couple false starts on it.

jb55:
thanks! testing again

jb55:
this is incredible. it reconnects instantly. THIS CHANGES EVERYTHING

jb55:
THANK YOU 0e94c48e26f448a01773ff02e397f014b6ff6449

PR #918, opened 2023-04-14
QR Code in profiles

ericholguin:
This PR adds a way to show other user's qr codes from their profile page when pressing the ellipsis button

qrcodes


jb55:
nice

PR #935, opened 2023-04-16
fix-repost-not-working-in-some-version

scoder1747:
fix-repost-not-working on some iOS versions


jb55:
this makes sense to me!

Issue #922, opened 2023-04-14
Multiple mentions break hyperlinks of @username (@npub display as #[3], #[4] etc.) [bug]


alltheseas:
_what happens_
Instead of names, or hyperlinks, only  #[3], #[4] is displayed

![image](https://user-images.githubusercontent.com/64376233/232161457-0f27f14d-1fc4-406b-b6e8-58ee1a3bd849.jpeg)

https://damus.io/note1rttctwwvhyclchzctc5f3d0076ynp97zdw2m59qlar24n2n6lufqn24u0z

_what I think should happen_
There should be no #[3], #[4] displayed in lieu of hyperlinked @ names

alltheseas:
> benweeks9: Suspect lack of spaces caused it.
> Tyiu: Funny thing is that I got the notification in Damus even though my mention doesn’t render

tyiu:
This looks fixed.

PR #936, opened 2023-04-16
[issues-992] Fix parse mention without space in content

joshuajiangdev:
Fix https://github.com/damus-io/damus/issues/922 , and add unit test.

Simulator Screenshot - iPhone 14 Pro - 2023-04-15 at 02 34 59


jb55:
thanks!

PR #939, opened 2023-04-16
Improve sats numeric entry for zaps

tyiu:
Changelog-Changed: Add number formatting for sats entry and use selected zaps amount from picker as placeholder
Changelog-Fixed: Do not allow non-numeric characters for sats amount and fix numeric entry for other number systems for all locales

Simulator Screen Recording - iPhone 14 Pro - 2023-04-16 at 18 13 03


jb55:
nice

PR #943, opened 2023-04-16
Do not translate own notes if logged in with private key

tyiu:
Changelog-Fixed: Do not translate own notes if logged in with private key


tyiu:
> why does privkey user matter ?

If you're not the private key user, it's possible you're just lurking on somebody's account and you might not speak their language, so you might need translations.

PR #944, opened 2023-04-17
Add setting to hide reactions

tyiu:
Changelog-Added: Add setting to hide reactions


jb55:
A better solution would be to not fetch reactions when this is enabled. Saves bandwidth.

alltheseas:
@tyiu does your x @jb55 's proposed solution also eliminate notifications from shaka/like?

tyiu:
> @tyiu does your x @jb55 's proposed solution also eliminate notifications from ~shaka/like~ reactions?

It should. The caching and filtering on the notifications tab isn't exactly the best, so some older events linger around. But generally yes.

PR #946, opened 2023-04-17
added nokyctranslate dropdown option to settings

symbsrcool:
Makes it easier for people to use nokyctranslate.com translation service by removing the need to enter a custom url via libtre translate, and instead new dropdown menu item for nokyctranslate.com with a button linking to the website. Verified that translations are working when this option is selected and API Key is given. 

I think I got everything and the merge should be easy, could one of you take a quick peek and let me know if I did something wrong? I don't do much swift development
@tyiu @jb55

### DIFF
--- a/damus/Models/TranslationService.swift
+++ b/damus/Models/TranslationService.swift
@@ -19,6 +19,7 @@ enum TranslationService: String, CaseIterable, Identifiable {
     case none
     case libretranslate
     case deepl
+    case nokyctranslate
 
     var model: Model {
         switch self {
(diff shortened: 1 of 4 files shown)


jb55:
very cool. what's the backend for this? are the translations good ?

symbsrcool:
Currently the backend is AWS Machine Translation. We’re working to get more cutting-edge models like NLLB (which supports 200 langauges) but the GPU server costs are expensive so we need to wait until a certain number of users are using the service.

Translations seem high quality but we need more community feedback and testing.

We launched a week ago so just getting started and will be adding lots more features. If you have any ideas/suggestions please let us know, our official Nostr npub is

npub109pkdu0k0dau2cz063lzrgn7dl9l7lk8u7njcm2v8pk4pl2a9uzqf6u4w5

PR #965, opened 2023-04-18
Add setting to hide reactions

### LINKED ISSUE #645: Add toggle: remove 🤙 (shaka/like) (OnlyZaps) [UX][Feature Request]
## User Story ##
As a low-time preference Damus user who wants to remove noise from my notes, I want an option to remove all 🤙 (shaka/like), so that I can only see the true signal  of value & feedback - ⚡ (zaps).

### Acceptance Criteria ###
1. In `settings` menu, user can toggle (turn off, or turn on) 🤙 (shaka/like)
2. If 🤙 (shaka/like) toggle is off, timeline no longer shows 🤙 (shaka/like)

Mockup
No Shaka Likes View

Settings Shaka View

Origin

"_`I wonder if removing likes/shaka/heart and keeping only ⚡️ will improve the time preference and quality of discourse`_"

"_`I've considered it but that would be a pretty big change and i think a lot of people would not like that`_"
-Will

"_`Could make it a personal setting where users decide to do away with seeing likes and only see zaps. Would be interesting to see if the user base chooses to move the experience in this direction.`_"
-Steven (GangGreen) npub1j9pu0navtt5ac96kmvqlwtjr2msy6egpa4xvr72hme3vatcys5gqknemxn

"_`Yeah its a good idea`_"
-Will

https://damus.io/note14h67g7n2x5ghs3kp0kz89ram8umz9guqvuzwdqv3v8v2wey8q8ss8sqluu

tyiu:
Changelog-Added: Add setting to hide reactions

Fixes #645

### DIFF
--- a/damus/ContentView.swift
+++ b/damus/ContentView.swift
@@ -467,13 +467,13 @@ struct ContentView: View {
             self.damus_state?.pool.connect_to_disconnected()
         }
         .onReceive(handle_notify(.new_mutes)) { notif in
-            home.filter_muted()
+            home.filter_events()
         }
         .onReceive(handle_notify(.mute_thread)) { notif in
-            home.filter_muted()
+            home.filter_events()
         }
         .onReceive(handle_notify(.unmute_thread)) { notif in
-            home.filter_muted()
+            home.filter_events()
         }
         .onReceive(handle_notify(.local_notification)) { notif in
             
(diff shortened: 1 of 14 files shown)


damus/Models/NotificationsModel.swift
```diff
@@ -144,12 +144,14 @@ class NotificationsModel: ObservableObject, ScrollQueue {
...
             notifs.append(.repost(evid, evgrp))
         }
-        
-        for el in reactions {
-            let evid = el.key
-            let evgrp = el.value
-            
-            notifs.append(.reaction(evid, evgrp))
+
+        if !damus_state.settings.hide_reactions {
```
jb55, line 148: I don't the we need to add this, or change the build_notifications or filter interface. If we're not fetching reactions then the reactions would be empty

tyiu: Fixed.

tyiu:
Merged in https://github.com/damus-io/damus/commit/66641fc9ae21e46ae4a7623ca237a27ca7a17f18

Issue #949, opened 2023-04-17
Hashtags in foreign languages 


lima46zulu:
Damus doesn't render hashtags in foreign languages properly. Same issue with accented characters. 

Many countries are using english hashtags, but most if not all popular hashtags in Japan are in Japanese. For example [写真好きな人と繋がりたい](https://twitter.com/hashtag/%E5%86%99%E7%9C%9F%E5%A5%BD%E3%81%8D%E3%81%AA%E4%BA%BA%E3%81%A8%E7%B9%8B%E3%81%8C%E3%82%8A%E3%81%9F%E3%81%84?src=hashtag_click).

Link to test note: https://damus.io/note10l0fzzkwfk45ckyc6swkyddeplc62xhqqxe4faln2c9g3fgv6ftqqvl06r

![IMG_5284](https://user-images.githubusercontent.com/109115349/232501949-c99913d8-0584-4938-b6e8-4e0722b663a9.jpeg)

alltheseas:
Thank you for submitting. This is in the backlog 🙏

garyray-k:
https://forums.swift.org/t/how-to-include-accented-characters-in-swift-url-without-percent-encoding/39215

garyray-k:
After digging into this for a bit, I think the bug lies somewhere in `damus.c`'s `parse_hashtag` function. Specifically the `consume_until_boundary` -> `is_boundary(c)` call. `is_boundary` uses `isalnum(c)` which does not account for symbols or non-English characters.

Noting these things here so I can draft a PR sometime later. My only hesitation is if this "boundary" checking is used elsewhere, there may be second/third order effects from changing this that are not accounted for yet.

alltheseas:
@lima46zulu can you correct issue title to correct spelling of `Hashtags`, so this ticket is searchable 🙏

velram:
Is anyone working on this issue? If not shall I pickup this issue?

velram:
> After digging into this for a bit, I think the bug lies somewhere in `damus.c`'s `parse_hashtag` function. Specifically the `consume_until_boundary` -> `is_boundary(c)` call. `is_boundary` uses `isalnum(c)` which does not account for symbols or non-English characters.
> 
> ~Noting these things here so I can draft a PR sometime later. My only hesitation is if this "boundary" checking is used elsewhere, there may be second/third order effects from changing this that are not accounted for yet.~
> 
> Edit: I lack the C experience/knowledge to complete this.

[...]

PR #970, opened 2023-04-18
₿itcoin B hashtag

garyray-k:
Not sure much explanation is needed. But my phone autocorrects "bitcoin"  to "₿itcoin" so this would be neat.

### DIFF
--- a/damus/Util/Hashtags.swift
+++ b/damus/Util/Hashtags.swift
@@ -30,6 +30,7 @@ struct CustomHashtag {
 let custom_hashtags: [String: CustomHashtag] = [
     "bitcoin": CustomHashtag.bitcoin,
     "btc": CustomHashtag.bitcoin,
+    "₿itcoin": CustomHashtag.bitcoin,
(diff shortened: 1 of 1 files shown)


jb55:
uhh wut lol

garyray-k:
I see now my half-ass attempt at a PR during fiat-mining lunch time was just that, half-ass.

But this might be related to https://github.com/damus-io/damus/issues/949 given the "non-English" character ₿ seemingly interrupting ₿itcoin from being a hashtag. Wondering if the same is happening with non-English languages too.

Full disclaimer, this is the first Swift I've ever written so I could be WAY off. 🤷‍♂️

garyray-k:
> Can you add a screenshot

FWIW, actually took a screenshot. Thank you for the suggestion.
Screenshot 2023-04-18 at 10 05 57 PM

alltheseas:
image

This is what damus looks like on my end. It looks clear and elegant. 

I don't think your PR will pass the Will test. Do you want me to suggest some other open issues you could work on?

PR #975, opened 2023-04-19
Update Custom Zap View

ericholguin:
This PR updates the custom zap view:

 

customzapdark   customzaplight


jb55:
so good!

damus/Views/Zaps/CustomizeZapView.swift
```diff
@@ -166,76 +306,52 @@ struct CustomizeZapView: View {
...
-                    .foregroundColor(.red)
+            AmountPicker
+            
+            ZapReply
+            
+            ZapButton
+            
+            Spacer()
+            
+            Spacer()
```
jb55, line 333: extra spacer here?

ericholguin: Pushed it up slightly, but not a huge difference so I removed it

jb55:
There doesn't appear to be a way to dismiss this view or the privacy selector view on macOS. Maybe we should have an X on the top left? Every sheet should probably have a close button like this. I always get reports of people who are confused and don't realize it's a sheet and they need to swipe down. on macOS it's even more confusing since you have to click outside of the view instead of swiping down.

jb55:
I find the amount input to be pretty confusing. I can't see if I've selected it or not.

jb55:
I'm going to merge this with a few adjustments and then we can iterate further

jb55:
merged in 8227be18736f1e1bdf5ca1ef6c60b93add62b31e with some tweaks from me in ea90fb042945beb9eabbb15e4ffe7d5f37f39591

PR #992, opened 2023-04-21
dm-sorted-based-on-last-event-created-date

scoder1747:
dm-sorted-based-on-last-event-created-date


jb55:
we already do insert sorting in handle_incoming_dms, if there might be a bug in there?

scoder1747:
> we already do insert sorting in handle_incoming_dms, if there might be a bug in there?

I saw issue where the users with latest dms weren't shown at the top. But I am not able to replicate now. I'll close this

PR #998, opened 2023-04-22
Fix crash with LibreTranslate server setting selection and remove delisted vern server

tyiu:
Broken after the settings refactor

Changelog-Fixed: Fix crash with LibreTranslate server setting selection and remove delisted vern server


jb55:
thanks!

PR #1007, opened 2023-04-24
Most recent conversations are not sorted to the top instantaneously

scoder1747:
This resolves the issue like below where William should have shown ahead of Swift because of the recency of conversation which was just 25 seconds ago comparing with 12 hours ago one.
IMG_7758


jb55:
thanks!

jb55:
I can't make sense of this change =/

PR #1012, opened 2023-04-25
Auto-fill-previous-zap-comment-for-the-same-zap-amount

scoder1747:
This pr is about persisting zap comment against the zap amount. If I zap an amount with a message, I will see that message as a comment next time when I select the same amount.

Demo video:
https://nostr.build/av/c3f300b038403d255e3e5fad4466cef887c3c9aeb7696542ff11f94c60d6dcc1.mov


jb55:
ah I'm not sure if I like this :(

Issue #1021, opened 2023-04-26
Attach my pub key to a NFC chip [feature request][conferences]


alltheseas:
## User Story
As a Damus user who participates in in person conferences, I want to be able to share my public key via NFC, so that others can easily open, and have the option to follow my pubkey.

### Acceptance Criteria
1. In Damus there is a straightforward method for user to attach their pub key to a NFC chip 
2. If user loads their npub to a NFC chip, when scanned by a NFC reader, scanning device should read & return the user's npub

demo vid: https://trello.com/1/cards/6445ed711078397bf9eca445/attachments/644936900ebdf9dd9f37a735/download/CleanShot_2023-04-25_at_23.56.26.mp4
origin: @BenGWeeks

BenGWeeks:
https://github.com/damus-io/damus/pull/1033

PR #1033, opened 2023-04-28
NFC writer for in-person zap badges

BenGWeeks:
This update adds the ability for users to write their nPubs to an NFC chip. This can be used in Miami for the in-person zapping badges. This uses a new NFCWriter class and a button added to the bottom of the left navigation bar.

@jb55 if you could add us to the App Store Connect team with a developer role so that I can create certs and provisioning rather than having to change the bundle identifier etc. in order to add NFC capability (check you don't amend lines 2000 onwards in project.pbxproj). damus.entitlements needs lines 5 + 7, but I had to remove domains etc. Happy to amend the PR if able to add.

I have noticed a bug @alltheseas - where the feed doesn't load if you go directly into Damus using a nostr:<nPub> link.


alltheseas:
Created bug when entering damus from nostr: link per your report @BenGWeeks https://github.com/damus-io/damus/issues/1062

BenGWeeks:
fixes https://github.com/damus-io/damus/issues/1021

BenGWeeks:
Instructions here:

https://damus.io/note1thn6satae7kdwp6ms0k25uapejtppfuyk4lc27r8zqsvur07gnfqhdddm4

NB Location of QR and NFC icons will move in redesign.

PR #1034, opened 2023-04-28
fixed-sats-plurality

scoder1747:
IMG_23142F790F54-1

### DIFF
--- a/damus/Views/Zaps/CustomizeZapView.swift
+++ b/damus/Views/Zaps/CustomizeZapView.swift
@@ -147,7 +147,7 @@ struct CustomizeZapView: View {
                    self.custom_amount_sats = nil
                }
             }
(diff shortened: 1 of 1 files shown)


jb55:
nice, thx

tyiu:
@scoder1747 @jb55 I'll keep saying it, but this doesn't work for all locales. We need to use Localizable.stringsdict. Arabic has 6 plural rules. I'll fix it.

tyiu:
This is the actual fix: https://github.com/damus-io/damus/pull/1048/commits/31f68c112a0ff52631a98b20cd2da8fe870449be

jb55:
Maybe we can make a localization plurality helper so these cases are easier ?

tyiu:
(I amended the commit I referenced to include tests)

> Maybe we can make a localization plurality helper so these cases are easier ?

Helper would be half the battle. The other half would involve adding some script to export the strings correctly to Localizable.stringsdict. But depending on the complexity of the pluralized string, this may or may not be very difficult to do programmatically.

PR #1047, opened 2023-04-29
Fix bug with reaction notifications referencing the wrong event

tyiu:
Changelog-Fixed: Fix bug with reaction notifications referencing the wrong event

NIP-25 says that it should be the last tag in the list, not the first.
https://github.com/nostr-protocol/nips/blob/master/25.md#tags


jb55:
Thanks!

Issue #1044, opened 2023-04-28
npub search fails on initial attempt, succeeds on subsequent attempt [bug]


alltheseas:
_what happens_
Sometimes when I search for a particular npub I have copied from elsewhere, on the first search attempt no results appear.
On a subsequent attempt npub appears.

https://user-images.githubusercontent.com/64376233/235260150-82c8c1e8-f66b-4c6a-95ea-16362b134f36.MOV

gladiusKatana:
Just found _one_ solution - will post PR momentarily.  (It's just a multithreading issue or at least that's one way to remove the bug - see below. Also - bug may not occur as often when searching larger profiles eg. @Damus [npub18m76awca3y37hkvuneavuw6pjj4525fw90necxmadrvjg0sdy6qsngq955]

In case anyone reading this wants to replicate the issue, Here is a fresh npub (will also post it in PR; searching this throwaway account should ~always~ result in the bug : npub1j6rzjsn0pgvf85cfa32glnx7djmuk6t5pz4c3yrly370z2cvfhussmdg8y

The fix: 

in `SearchingEventView` / `handle_search` : 

```swift
        case .profile:
            find_event(state: state, evid: evid, search_type: search_type, find_from: nil) { _ in
                DispatchQueue.main.asyncAfter(deadline: .now() + 0.1 ) { // <-----
                    if state.profiles.lookup(id: evid) != nil {
                        self.search_state = .found_profile(evid)
                        return
                    } else {
                        self.search_state = .not_found
                    }
                } //<-----
            }
```

PR #1051, opened 2023-04-30
Fix Issue #1044 : "npub search fails on initial attempt, succeeds on subsequent attempt"

gladiusKatana:
-Fix issue #1044

#### Change:
-Wrap the `find_event(...)` callback in the `.profile` case in a dispatch to the main queue w/ a slight time delay

#### Findings:
-Time delay value =0 does not fix
-To replicate bug, needed to use a new npub (eg my account created for testing, `npub1j6rzjsn0pgvf85cfa32glnx7djmuk6t5pz4c3yrly370z2cvfhussmdg8y`)

 #### Demo

Before:

https://user-images.githubusercontent.com/47217795/235339653-854e2d7e-a0bc-47ca-a98d-cb631f654bc5.mov

After:

https://user-images.githubusercontent.com/47217795/235339659-0b202b48-faf7-4381-bbc9-9b7936030f44.mov

### DIFF
--- a/damus/Views/Search/SearchingEventView.swift
+++ b/damus/Views/Search/SearchingEventView.swift
@@ -93,11 +93,13 @@ struct SearchingEventView: View {
             }
         case .profile:
             find_event(state: state, evid: evid, search_type: search_type, find_from: nil) { _ in
-                if state.profiles.lookup(id: evid) != nil {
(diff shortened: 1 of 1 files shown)


jb55:
this makes no sense to me but that's pretty standard with swiftui

PR #1060, opened 2023-05-01
Changelog-Fixed: Save Jacks soul

BenGWeeks:
Very minor amendment to stop the bouncing of the bolt symbol as you click it.

If particularly want it to switch to a horizontal bolt (I personally prefer keeping it vertical) you could probably also resolve using .frame().


jb55:
Haha thank you

PR #1064, opened 2023-05-02
fix issue where new events don't show when the app resumes

bryanmontz:
This is an attempt to fix the issue where new events don't always load when Damus is resumed. Before this change, you could change the main tab and go back to the timeline view and events would load as expected. After the change, you shouldn't have to do that.

### DIFF
--- a/damus/Views/TimelineView.swift
+++ b/damus/Views/TimelineView.swift
@@ -17,6 +17,7 @@ struct TimelineView: View {
     @Binding var loading: Bool
     @State var offset = CGFloat.zero
     
+    @Environment(\.scenePhase) var scenePhase
     @Environment(\.colorScheme) var colorScheme
 
     let damus: DamusState
@@ -59,6 +60,16 @@ struct TimelineView: View {
         }
         .onAppear {
             events.flush()
+        }
+        .onChange(of: scenePhase) { phase in
+            switch phase {
+            case .active:
+                events.flush()
+            case .background, .inactive:
+                break
+            @unknown default:
+                break
+            }
         }
     }
 }


jb55:
not sure if flushing is the issue, because clicking the notification tab should flush events. clicking the tab does nothing, the only thing I've been able to do to get it to reconnect is to goto global feed to force network activity?

bryanmontz:
Yeah, I wasn't sure about it and haven't gotten to pursue it further yet.

bryanmontz:
I tested this a bunch more, and I believe that this either fixes the issue entirely or at least makes it a lot better.

I can reproduce the issue with the existing code like this:
Launch app in simulator with my "real" account, let it update with events and settle.
Home out of the app in the simulator.
Wait at least a minute or two.
In a different simulator or on device, from a different account that the real account follows, post a test message.
Resume app in simulator with real account.
Scroll to top, note that the new post is not there.
Switch to Notifications tab (`EventHolder.flush()` is not called yet)
Switch back to Home tab, `flush()` is now called and the post is visible.

With the fix in place, and the same steps as above:
`flush()` gets called immediately on resume and the post is visible.

jb55:
Even if this is a some-views-not-getting-refreshed-when-resuming issue we don’t want to flush events from a scene state change like this. The purpose of flush is to flush events that have already been received over the wire and are ready to be put into the timeline.

if the app was just coming out of the background it wouldn’t have any events to flush, it first needs to reconnect, receive the events and queue them into the HomeModel EventHolders and the NotificationsModel event holders. This would light up the notification dots which we do not see when we hit this bug.

Even if this somehow fixes it which I have yet to see, we don’t want to flush on resume, we want to flush on a user action such as when they click the notification tab. The home and notifications tab should have the notification dot when it has unflushed events, it doesn’t even have that. Flushing does nothing if there are no new events to flush. Switching home and notification tabs always triggers a flush as well, when hitting this bug switching between home and notifications never flushes, the only way to fix it is to go to the universe tab.

From what I can tell, the real issue that the websocket connection thinks it’s still connected when it’s not and does not try to reconnect after resume, and which is why when you go to the universe tab it tries to query data but can’t, so it reconnects and then finally we get all the data.

maybe I’m wrong, but I can’t see how flushing fixes the real issue some users are experiencing 🤔 unless there are two separate bugs and this is attempting to fix a different one.

jb55:
Damus had this bug with the Starscream code as well (isConnected was true when it wasn’t), my hack was to test the connection every 5 seconds or so and reconnect if it was stale. maybe we need damus to ping on resume or something 😔

bryanmontz:
I think you're right because I saw another symptom of this: on the RelayConfigView, the relays all report that they are connected immediately when you resume the app after a long break. I wonder if manually pinging all the relays on app resume would help. The ping would probably fail and cause the reconnect logic to hit. I can try this later.

bryanmontz:
And thanks for that explanation. Very helpful.

bryanmontz:
Lol, sorry I suggested what you had already suggested. Somehow I missed that.

PR #1080, opened 2023-05-05
Iconography and hashtag icons for the new Damus redesign.

robagreda:
Added iconography for the new damus re-design, as well added new hashtag icons and replaced the old ones keeping a square frame.


BenGWeeks:
Nice. Does this mean I should start implementing the new iconography @jb55 ?! :-)

PR #1082, opened 2023-05-05
cache user profiles in CoreData for quick lookup after app restarts

bryanmontz:
This PR implements a CoreData-based caching layer for user profiles. This improves the problem of pubkeys beings displayed in feeds, when you've already visited a profile during a previous app session.

Both of these videos were taken immediately after launching the app. For the After video, I had visited several of the profiles in the feed in a previous app session in order for them to be cached in the database.

|Before|After|
|---|---|
|Notice pubkeys in feed|No pubkeys in feed (visited)|
|bad|good|


jb55:
👀

PR #1099, opened 2023-05-05
Chanelog-Fixed: Crash loading artifacts

joelklabo:


bryanmontz:
Looks like a good fix to me. Those explicit unwraps will haunt you.

bryanmontz:
Though I like your changed code better than the original, Will is right that this is probably not going to fix a crash.

PR #1100, opened 2023-05-06
fix: npub search fails on the first try

bryanmontz:
This PR fixes the issue where npub search fails on the first try.

What was happening:
`SearchEventView` calls `find_event(state:evid:search_type:find_from:callback:)` to search for the entered npub. The search actually succeeded, but the call site was ignoring the returned event and never caching the profile. The reason it works the second time is because the `RelayPool` eventually got around to caching the profile. This is also why the other PR related to this issue works by adding a short delay.

How it's fixed:
Now `SearchingEventView` will call `process_metadata_event(events:our_pubkey:profiles:ev:)`, and upon completion, will update the view state with the successful result.

### DIFF
--- a/damus/Models/HomeModel.swift
+++ b/damus/Models/HomeModel.swift
@@ -761,10 +761,11 @@ func guard_valid_event(events: EventCache, ev: NostrEvent, callback: @escaping (
     }
 }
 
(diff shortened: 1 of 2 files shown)


bryanmontz:
If you don't like the view calling `process_metadata_event`, we could instead listen for the `profile_updated` notification and check if the updated profile matches the npub we were searching for.

PR #1101, opened 2023-05-06
fix posts with no uploadedURLs always get two blank spaces at the end

bryanmontz:
In the case where a new post does not contain `uploadedMedias`, the post will always get two blank spaces appended at the end.

damus_—_PostView_swift

### DIFF
--- a/damus/Views/PostView.swift
+++ b/damus/Views/PostView.swift
@@ -84,8 +84,10 @@ struct PostView: View {
         let imagesString = uploadedMedias.map { $0.uploadedURL.absoluteString }.joined(separator: " ")
         
         let img_meta_tags = uploadedMedias.compactMap { $0.metadata?.to_tag() }
-
-        content.append(" " + imagesString + " ")
+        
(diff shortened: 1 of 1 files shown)


tyiu (approved):

scoder1747 (approved):

damus/Views/PostView.swift
```diff
@@ -84,8 +84,10 @@ struct PostView: View {
         let imagesString = uploadedMedias.map { $0.uploadedURL.absoluteString }.joined(separator: " ")
         
         let img_meta_tags = uploadedMedias.compactMap { $0.metadata?.to_tag() }
-
-        content.append(" " + imagesString + " ")
+        
+        if !imagesString.isEmpty {
```
scoder1747, line 88: Another solution would be to keep this line content.append(" " + imagesString + " ") before line#82 which is var content = self.post.string.trimmingCharacters(in: CharacterSet.whitespacesAndNewlines)

Issue #1038, opened 2023-04-28
User tag autocomplete message area blocks text input


dmnyc:
When attempting to autocomplete a username, the results overlay blocks the original message entry area. If a username is not found, it is very difficult to get back to the message field. Requires hitting the delete key many times until the npub is deleted completely. 

I suggest allowing more space for this so the user can correct mistakes in thier note entry field.
![IMG_5890](https://user-images.githubusercontent.com/3638467/235214366-7e07a036-eb85-4536-8b0c-062a0d6ed484.jpg)

alltheseas:
Great suggestion. Recreated bug. Adding to backlog. https://user-images.githubusercontent.com/64376233/235231630-c36aaa9d-4e41-4e5e-8643-73e5c1b8bf92.MOV

alltheseas:
@dmnyc is this resolved for you?

dmnyc:
Yes, the display box size and position is no longer an issue, however another tagging issue appeared recently. 
https://github.com/damus-io/damus/issues/1544

PR #1111, opened 2023-05-09
Fix Issue #1038 - "User tag autocomplete message area blocks text input"

gladiusKatana:
Fix Issue #1038

 Fix:
 -Cursor in a post/reply now remains visible (regardless of initial location) even when search/autocomplete view is present

 Change:
 -Conditionally set the `.isScrollEnabled` in `TextViewWrapper` (ie, default true; & false only when `UserSearch` is present)

Demo:

https://github.com/damus-io/damus/assets/47217795/374e298c-0f81-4306-94c1-97244d0e4cd5


jb55:
thanks!

joelklabo:
LGTM!

jb55:
cherry-picked into f9a572faa28422a4df30af98b26da155df96a0d0

PR #1113, opened 2023-05-10
Add translate.nostr.wine to available translation services

nostr-wine:
Added translate.nostr.wine to the translation service options in Settings. Thank you for making it very easy. Please let us know if we missed anything. 

Powered by DeepL and backed up by Google Cloud Translation API if the language is not supported.


nostr-wine:
Up to you if you want to replicate the translateWith block. We could alternatively just use the LibreTranslate one.

PR #1122, opened 2023-05-11
Translations

tyiu:


damus/Views/Buttons/AlbyButton.swift
```diff
@@ -23,7 +23,7 @@ struct AlbyButton: View {
             HStack {
                 Image("alby")
                 
-                Text("Connect to Alby")
```
jb55, line 26: ya'll are too fast. I've already renamed this to `Attach Alby Wallet`

tyiu: No, you're too fast! 😄 It's all good, rebased and added another commit to export strings for translation.

PR #1187, opened 2023-05-27
Setup, Login, Create Account redesign

ericholguin:
Implementing @robagreda designs from here: [Figma](https://www.figma.com/file/ORaT1T0Ywfbm0sIjwy5Rgq/Damus-iOS?type=design&node-id=1069-28898&t=kGXqjuQOm7hSXbzW-0)

**Account Creation flow**:

----------**iPhone 14 Pro Max iOS 16**-------------------**iPhone 12 Mini iOS 15.5**----------
iPhone 14 Pro Max  iPhone 12 mini


jb55:
looking at this now

jb55:
One thing I noticed is that it sets the display_name (an unspec'd field)
instead of `name`. We probably want to be setting name instead of we are
pushing users to a single username.

jb55:
merged!

PR #1188, opened 2023-05-27
Beautify iconography

BenGWeeks:
This PR uses the iconography from the Figma design system. There are a couple more icons to add (see comments in the Design System page).

CleanShot 2023-05-27 at 04 33 33

CleanShot 2023-05-27 at 04 33 50

CleanShot 2023-05-27 at 04 34 07

CleanShot 2023-05-27 at 04 35 00

CleanShot 2023-05-27 at 04 35 29

Will do branding and colours in a separate PR.


bryanmontz:
@BenGWeeks Looks good! Couple comments:
1. This is a *lot* of icons. They are not all used, which adds bloat to the binary. Maybe we could only add the images that are actually in use?
2. It would be nice if the screenshots you provide in the PR were in a markdown table with "Before" on the left and "After" on the right, so that reviewers can see the changes more easily. https://markdown.land/markdown-table
3. All the new icons are specified in the assets file with 1x, 2x, and 3x scales, but there's actually only one file that they're pointing to. Instead, you can mark the asset as "Single Scale":
damus_—_Assets_xcassets_and_bryanmontz_Bryan___Development_Projects_damus-ben  

4. Will has previously indicated that he might prefer 3 sizes of pngs rather than one svg, because svgs require a small amount of processing time. Not sure if that's still the case or if it applies here.

BenGWeeks:
NB More icons being uploaded to the repo soon from @robagreda after which there will be another update in a separate PR.

BenGWeeks:
> @BenGWeeks Looks good! Couple comments:
> 
> 1. This is a *lot* of icons. They are not all used, which adds bloat to the binary. Maybe we could only add the images that are actually in use?

Adding the icons was actually a separate PR from @robagreda. However, do the open source nature of the project, might be worth making the icons available so developers know they can use them without resorting to custom creating or SF Symbols. We _could_ have them in the directory but not added to the project perhaps? (Will devs think to look?)

jb55:
On Mon, May 29, 2023 at 05:03:01AM -0700, Bryan Montz wrote:
>@BenGWeeks Looks good! Couple comments:
>1. This is a *lot* of icons. They are not all used, which adds bloat to the binary. Maybe we could only add the images that are actually in use?

Agreed

>3. All the new icons are specified in the assets file with 1x, 2x, and 3x scales, but there's actually only one file that they're pointing to. Instead, you can mark the asset as "Single Scale":

I assumed this converted them to many differet sizes for different
devices using the same source image. Does single scale do what we want?

>4. Will has previously indicated that he might prefer 3 sizes of pngs rather than one svg, because svgs require a small amount of processing time. Not sure if that's still the case or if it applies here.

Again, I'm working off the assumption that it is creating appropriate
per-device sizes based on the source image, avoiding processing time at
runtime. This is ideal if it saves on startup time. I'm assuming lots of
things here though and not sure how it all works.

Cheers,
Will

jb55:
On Mon, May 29, 2023 at 10:42:09AM -0700, Ben Weeks wrote:
>> @BenGWeeks Looks good! Couple comments:
>>
>> 1. This is a *lot* of icons. They are not all used, which adds bloat to the binary. Maybe we could only add the images that are actually in use?
>
>Adding the icons was actually a separate PR from @robagreda. However, do the open source nature of the project, might be worth making the icons available so developers know they can use them without resorting to custom creating or SF Symbols. We _could_ have them in the directory but not added to the project perhaps? (Will devs think to look?)

We should test to see if they are even packaged until used? That would
be a nice optimization. Maybe there could be some kind of AssetsPreview
thing that could be there for devs but not actually added to the
resulting app. Then you would have to copy them over manually to Assets
when you want to use them.

BenGWeeks:
Any easy way to check whats's packaged other than file size?

BenGWeeks:
image

Assuming it's correct!

BenGWeeks:
image

jb55:
On Mon, May 29, 2023 at 01:37:00PM -0700, Ben Weeks wrote:
>image
>
>Assuming it's correct!

good to know!

jb55:
Pushed to redesign branch!

To github.com:damus-io/damus
    bf01d08e..ed4ef0e2  redesign -> redesign

PR #1192, opened 2023-05-27
Beautify Fonts: Updated UX to use custom font

BenGWeeks:
This PR updates the fonts to Inter as per Figma.

This also uses a custom Font Manager that extends Font so that existing code doesn't need to be changed and we can continue to use .body, .caption, etc.

Will tweak font sizes most likely when get to rolling-out the full re-design.


BenGWeeks:
NB More icons being uploaded to the repo soon from @robagreda after which there will be another update in a separate PR.

jb55:
Love Inter!

BenGWeeks:
Me too - although more lines of code than I would have expected to change the font!!!

BenGWeeks:
FYI I'll be tweaking the font sizes/weights more during implementation of the new UX.

jb55:
cherry-picked into the redesign branch:

1c4339f0f1eae992f79ed8c148efacd1ace4f9b6

thanks!

PR #1197, opened 2023-05-29
Advanced recommended relays (suggest others relays)

BenGWeeks:
This pull request provides recommended relays in the Relays area based on relays the people you follow.

This is draft, as I hit an issue where profile.contacts was nil and wasn't sure why. I was referencing the implementation in ProfileView to instantiate FollowingModel.

CleanShot 2023-05-29 at 02 05 42

### DIFF
--- a/damus/ContentView.swift
+++ b/damus/ContentView.swift
@@ -282,8 +282,10 @@ struct ContentView: View {
                                 
                                 ToolbarItem(placement: .navigationBarTrailing) {
                                     HStack(alignment: .center) {
-                                        SignalView(state: damus_state!, signal: home.signal)
-                                        
+                                        if let pk = self.active_profile {
+                                            let profile_model = ProfileModel(pubkey: pk, damus: damus_state!)
+                                            SignalView(state: damus_state!, signal: home.signal, profile: profile_model)
+                                        }
                                         // maybe expand this to other timelines in the future
                                         if selected_timeline == .search {
                                             Button(action: {
--- a/damus/Models/ProfileModel.swift
+++ b/damus/Models/ProfileModel.swift
@@ -32,8 +32,8 @@ class ProfileModel: ObservableObject, Equatable {
     func get_recommended_relays() -> [(RelayDescriptor, Int)] {
         var recommended_relays: [(RelayDescriptor, Int)] = []
         
-        if let contact = self.contacts {
-            let contacts_pubkeys = contact.referenced_pubkeys.map { $0.ref_id }
+        if let contacts {
+            let contacts_pubkeys = contacts.referenced_pubkeys.map { $0.ref_id }
             //let following = FollowingModel(damus_state: damus, contacts: contacts)
             
             for pubkey in contacts_pubkeys.reversed() {
@@ -55,6 +55,22 @@ class ProfileModel: ObservableObject, Equatable {
         }
         
         return recommended_relays
+    }
+    
+    func fetch_relays_for_contacts() {
+        guard let contacts = self.contacts else {
+            return
+        }
+        
+        let relays_subid = UUID().uuidString
+
+        let contactsPubkeys = contacts.referenced_pubkeys.map { $0.ref_id }
+        
+        // Create a filter for relay profiles (kind 3) of all your follows
+        let filter = NostrFilter(kinds: [3], authors: contactsPubkeys)
+        
+        // Subscribe to the relays with the filter and handle the received events
+        damus.pool.subscribe(sub_id: relays_subid, filters: [filter], handler: handle_event)
     }
     
     func follows(pubkey: String) -> Bool {
--- a/damus/Models/ProfileModel.swift
+++ b/damus/Models/ProfileModel.swift
@@ -29,6 +29,34 @@ class ProfileModel: ObservableObject, Equatable {
         })
     }
     
+    func get_recommended_relays() -> [(RelayDescriptor, Int)] {
+        var recommended_relays: [(RelayDescriptor, Int)] = []
+        
+        if let contact = self.contacts {
+            let contacts_pubkeys = contact.referenced_pubkeys.map { $0.ref_id }
+            //let following = FollowingModel(damus_state: damus, contacts: contacts)
+            
+            for pubkey in contacts_pubkeys.reversed() {
+                let contact = ProfileModel(pubkey: pubkey, damus: damus)
+                if let relays = contact.relays {
+                    for (urlString, relayInfo) in relays {
+                        if let relayURL = RelayURL(urlString) {
+                            let newRelay = RelayDescriptor(url: relayURL, info: relayInfo)
+                            
+                            if let index = recommended_relays.firstIndex(where: { $0.0.url == newRelay.url }) {
+                                recommended_relays[index].1 += 1
+                            } else {
+                                recommended_relays.append((newRelay, 1))
+                            }
+                        }
+                    }
+                }
+            }
+        }
+        
+        return recommended_relays
+    }
+    
     func follows(pubkey: String) -> Bool {
         guard let contacts = self.contacts else {
             return false
@@ -151,7 +179,6 @@ class ProfileModel: ObservableObject, Equatable {
         }
     }
 }
-
 
 func count_pubkeys(_ tags: [[String]]) -> Int {
     var c: Int = 0
--- a/damus/Views/Relays/RelayConfigView.swift
+++ b/damus/Views/Relays/RelayConfigView.swift
@@ -141,6 +141,13 @@ struct RelayConfigView: View {
                     .padding(.bottom, 5)
             }
         }
+        .onAppear() {
+            profile.subscribe()
+            profile.fetch_relays_for_contacts()
+        }
+        .onDisappear {
+            profile.unsubscribe()
+        }
         .navigationTitle(NSLocalizedString("Relays", comment: "Title of relays view"))
         .navigationBarTitleDisplayMode(.large)
         .toolbar {
--- a/damus/Views/Relays/RelayConfigView.swift
+++ b/damus/Views/Relays/RelayConfigView.swift
@@ -9,24 +9,20 @@ import SwiftUI
 
 struct RelayConfigView: View {
     let state: DamusState
+    
     @State var new_relay: String = ""
     @State var relays: [RelayDescriptor]
     @State private var showActionButtons = false
     
+    @StateObject var profile: ProfileModel
+    
     @Environment(\.dismiss) var dismiss
     
-    init(state: DamusState) {
+    init(state: DamusState, profile: ProfileModel) {
         self.state = state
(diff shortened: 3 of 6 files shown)


BenGWeeks:
@alltheseas This will mean that the PRs for people requesting relays to be added will be negated by this improved functionality as relays will be recommended on merit rather than using a default list (we could also use similar functionality in the sign-up process).

damus/ContentView.swift
```diff
@@ -282,7 +282,10 @@ struct ContentView: View {
                                 
                                 ToolbarItem(placement: .navigationBarTrailing) {
                                     HStack(alignment: .center) {
-                                        SignalView(state: damus_state!, signal: home.signal)
+                                        if let pk = self.active_profile {
```
bryanmontz, line 285: You don't need the `self.` here, and you can use the new abbreviated syntax:
```swift
if let active_profile {
    ...
}
```

damus/Views/Relays/RelayConfigView.swift
```diff
@@ -9,24 +9,55 @@ import SwiftUI
...
+        self._profile = StateObject(wrappedValue: profile)
     }
     
-    var recommended: [RelayDescriptor] {
-        let rs: [RelayDescriptor] = []
-        return BOOTSTRAP_RELAYS.reduce(into: rs) { xs, x in
-            if state.pool.get_relay(x) == nil, let url = RelayURL(x) {
-                xs.append(RelayDescriptor(url: url, info: .rw))
+    var recommended: [(RelayDescriptor, Int)] {
+        //var suggested_relays: [RelayDescriptor] = []
```
bryanmontz, line 28: Remove commented line

damus/Views/Relays/RelayConfigView.swift
```diff
@@ -9,24 +9,55 @@ import SwiftUI
...
+        var suggested_relays: [(RelayDescriptor, Int)] = [] // The Int is to keep count of how many people we follow us this relay.
```
bryanmontz, line 29: "people we follow us this relay" -> "people we follow use this relay"

damus/Views/Relays/RelayConfigView.swift
```diff
@@ -9,24 +9,55 @@ import SwiftUI
...
+                            }
+                        }
+                    }
+                }
             }
         }
+        
+        return suggested_relays // j55: We should probably limit this to just 10 or so?!
+        
+        //let rs: [RelayDescriptor] = []
```
bryanmontz, line 55: Remove commented code

damus/Views/Relays/RelayConfigView.swift
bryanmontz, line 29: Keep terminology consistent when possible. Is it "recommended" or "suggested"?

damus/Views/Relays/RelayConfigView.swift
```diff
@@ -9,24 +9,55 @@ import SwiftUI
...
-        return BOOTSTRAP_RELAYS.reduce(into: rs) { xs, x in
-            if state.pool.get_relay(x) == nil, let url = RelayURL(x) {
-                xs.append(RelayDescriptor(url: url, info: .rw))
+    var recommended: [(RelayDescriptor, Int)] {
+        //var suggested_relays: [RelayDescriptor] = []
+        var suggested_relays: [(RelayDescriptor, Int)] = [] // The Int is to keep count of how many people we follow us this relay.
+        
+        if let profile_contacts = profile.contacts {
+            let contacts_pubkeys = profile_contacts.referenced_pubkeys.map { $0.ref_id }
+            let following = FollowingModel(damus_state: state, contacts: contacts_pubkeys)
```
bryanmontz, line 33: Why do you need the `FollowingModel`? It looks like you're just passing in pubkeys and then reading the same ones back out?

So maybe just:
```swift
for pubkey in contacts_pubkeys { ... }
```

damus/Views/Relays/RelayConfigView.swift
```diff
@@ -9,24 +9,55 @@ import SwiftUI
...
         _relays = State(initialValue: state.pool.our_descriptors)
+        self._profile = StateObject(wrappedValue: profile)
     }
     
-    var recommended: [RelayDescriptor] {
-        let rs: [RelayDescriptor] = []
-        return BOOTSTRAP_RELAYS.reduce(into: rs) { xs, x in
-            if state.pool.get_relay(x) == nil, let url = RelayURL(x) {
-                xs.append(RelayDescriptor(url: url, info: .rw))
+    var recommended: [(RelayDescriptor, Int)] {
```
bryanmontz, line 27: IMO this is too much "smarts" for a view to have. This kind of calculation should be done in a model or helper function, where it can be easily tested. In this case I think `ProfileModel` would be a better place for it.

damus/Views/Relays/RelayConfigView.swift
```diff
@@ -9,24 +9,55 @@ import SwiftUI
...
+        //var suggested_relays: [RelayDescriptor] = []
+        var suggested_relays: [(RelayDescriptor, Int)] = [] // The Int is to keep count of how many people we follow us this relay.
+        
+        if let profile_contacts = profile.contacts {
+            let contacts_pubkeys = profile_contacts.referenced_pubkeys.map { $0.ref_id }
+            let following = FollowingModel(damus_state: state, contacts: contacts_pubkeys)
+            
+            for pubkey in following.contacts {
+                let contact = ProfileView(damus_state: state, pubkey: pubkey)
```
bryanmontz, line 36: Creating a view just to access data and not actually show it is wasteful, and I get a runtime warning: "Accessing StateObject's object without being installed on a View. This will create a new instance each time."
damus_—_RelayConfigView_swift

It looks like all you want is the `ProfileModel` out of the view, so get those this way:
```swift
let profile = ProfileModel(pubkey: pubkey, damus: state)
```

damus/Views/Relays/RelayConfigView.swift
```diff
@@ -130,16 +172,18 @@ struct RelayConfigView: View {
                 }
             }
             
-            if recommended.count > 0 {
-                Section {
-                    List(recommended, id: \.url) { r in
-                        RecommendedRelayView(damus: state, relay: r.url.id, showActionButtons: $showActionButtons)
+            Section {
+                if recommended.count > 0 {
```
bryanmontz, line 176: It's better practice to use this instead, because the full count of the array isn't needed:
```swift
if !recommended.isEmpty { ... }
```

damus/Views/Relays/RelayConfigView.swift
```diff
@@ -41,6 +72,17 @@ struct RelayConfigView: View {
     
     var MainContent: some View {
         Form {
+            // jb55: profile.contacts is nil below. Not sure why.
```
bryanmontz, line 75: If you go to your Following from your profile, and then you go to the Relays view, this will cause a list of your following to show up at the top of the view:

iPhone_14_Pro

damus/ContentView.swift
BenGWeeks: Amended.

damus/Views/Relays/RelayConfigView.swift
BenGWeeks: Done.

damus/Views/Relays/RelayConfigView.swift
BenGWeeks: Now implements `func get_recommended_relays() -> [(RelayDescriptor, Int)] {}`

BenGWeeks:
Pushed latest changes. Still draft.

I think this may be causing "rate limited" warnings and contacts is still empty but getting late here.

damus/Views/Relays/RelayConfigView.swift
BenGWeeks: Removed.

damus/ContentView.swift
BenGWeeks: Put this back in as I thought it was causing a build error. May not have been actually.

bryanmontz: No build error. And you don't event need a separate named entity if you do it like this:
```swift
if let active_profile {
    let profile_model = ProfileModel(pubkey: active_profile, damus: damus_state!)
    SignalView(state: damus_state!, signal: home.signal, profile: profile_model)
}
```

damus/Models/ProfileModel.swift
```diff
@@ -29,6 +29,34 @@ class ProfileModel: ObservableObject, Equatable {
         })
     }
     
+    func get_recommended_relays() -> [(RelayDescriptor, Int)] {
+        var recommended_relays: [(RelayDescriptor, Int)] = []
+        
+        if let contact = self.contacts {
```
bryanmontz, line 35: You can do `if let contacts { ... }`

damus/Models/ProfileModel.swift
```diff
@@ -29,6 +29,34 @@ class ProfileModel: ObservableObject, Equatable {
...
+            let contacts_pubkeys = contact.referenced_pubkeys.map { $0.ref_id }
```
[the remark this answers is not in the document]
bryanmontz: Something like this:
```swift
private var relays_subid = UUID().uuidString
...
let filter = NostrFilter(kinds: [3], authors: contacts_pubkeys)
damus.pool.subscribe(sub_id: relays_subid, filters: [filter], handler: handle_event)
```

And then that `handle_event` will receive all of the kind 3 contacts events for all your follows. Gather those up and perform this operation.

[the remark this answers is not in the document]
BenGWeeks: Where does this go?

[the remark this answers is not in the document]
BenGWeeks: > ```swift
> private var relays_subid = UUID().uuidString
> ...
> let filter = NostrFilter(kinds: [3], authors: contacts_pubkeys)
> damus.pool.subscribe(sub_id: relays_subid, filters: [filter], handler: handle_event)
> ```

Where does this go?

damus/Models/ProfileModel.swift
BenGWeeks: Done.

damus/Models/ProfileModel.swift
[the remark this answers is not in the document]
BenGWeeks: Have added `profile.fetch_relays_for_contacts()` to RelayConfigView and the following function:

    func fetch_relays_for_contacts() {
        guard let contacts = self.contacts else {
            return
        }
        
        let relays_subid = UUID().uuidString

        let contactsPubkeys = contacts.referenced_pubkeys.map { $0.ref_id }
        
        // Create a filter for relay profiles (kind 3) of all your follows
        let filter = NostrFilter(kinds: [3], authors: contactsPubkeys)
        
        // Subscribe to the relays with the filter and handle the received events
        damus.pool.subscribe(sub_id: relays_subid, filters: [filter], handler: handle_event)
    }

Issue #1189, opened 2023-05-27
Global throttling events to one per pubkey (bounty!)


brugeman:
Global tab is only showing one event per pubkey, as written in the SearchHomeModel.

That might be a useful strategy for the raw global feed to limit the impact of a single spammer. However, it seems to do more harm than good for cases where global is limited by the user to some set of filtering relays - like paid relays, or curated feeds (i.e. wss://feeds.nostr.band/memes). 

I'd suggest adding a switch to turn-off this throttling, probably in the same screen where global relays are switched on and off.

brugeman:
Adding a 100k sats bounty for a solution to this.

alltheseas:
How do you make sure there are only memes on the meme relay?

brugeman:
> How do you make sure there are only memes on the meme relay?

That feed is managed by a person, the list of seeds is here https://feeds.nostr.band

PR #1198, opened 2023-05-29
Multiple nPub posts in Universe (Resolve Issue 1189)

BenGWeeks:
- Added setting to switch multiple posts per person on/off in a new settings screen (default is current behaviour)
- Added logic to switch off/on multiple posts per person in Search/Universe


BenGWeeks:
fixes #1189

Issue #991, opened 2023-04-21
Notify me if someone follows me setting


alltheseas:
## User Story

As a Damus user who enjoys interacting with other nostriches, I would like to be notified if a nostrich that is not following me follows me, so that I can learn more about them, and how they ran across my nostr profile.

### Acceptance Criteria
1. Damus user has a setting to turn on notifications if someone who is not following them follows them
2. This setting can be turned off or on at any time
3. Default is off
4. If setting is turned on, and a new follower follows said user, user receives a notification

_context_
See https://nostrgraph.net/dashboard

Origin: SerSleepy

alltheseas:
Https://damus.io/note1cja0z8jvt9kdkvn754m62w8muly8svc82mf3ratyp0affvjnknhsnnzm8x

![image](https://github.com/user-attachments/assets/127011e1-1624-40d8-89b2-0526f698f9fd)

> Now that you’ve got push notifications on Damus TestFlight, how about notifications for when someone follows? That’s basically the only reason I still open up Primal anymore.

alltheseas:
@jb55 advises this is a risky spam vector

alltheseas:
![image](https://github.com/user-attachments/assets/cb71674b-27c6-4229-9817-5261d414c814)

Customer feedback

https://damus.io/nevent1qqsgsevnw5khlw846ex6lmp5nh27lv7k95g4rkkyezzzup982wg4uwspzpmhxue69uhkummnw3ezuamfdejsz9rhwden5te0wfjkccte9ekxuct49ehx2aqpzemhxue69uhhyetvv9ujuurjd9kkzmpwdejhgqgmwaehxw309aex2mrp0yhxg6t9vahhjet8wfhhxtnrdakswywq5y

alltheseas:
Received feedback from customers onboarding that they expect to be notified if someone follows me when they are getting started. Context is in person onboarding.

alltheseas:
@scoder1747

alltheseas:
Closing as part of a backlog triage — we're narrowing focus to the most actionable issues to reduce noise for contributors.

If this is still relevant, please reopen with updated context or reproduction steps. Thanks for reporting!

PR #1200, opened 2023-05-29
Follow notifications (Resolve issue 991)

BenGWeeks:
ChangeLog-Added: Ability to receive follow notifications (default setting is off) (closes damus-io/damus#991)

Will set this to draft as for some reason the notification is not appearing - think there's a filter somewhere I didn't spot. If we like this proposed change will spend more time to see if I can get to the bottom of it, or anyone more familiar with notifications?


jb55:
It's not as simple as this unfortunately. This would notify you anytime
someone updates their contact list with you on it. To do proper follow
notifications we'll need to keep track of the first time you were added
to a contact list and no future times.

BenGWeeks:
Aaah! That sounds a little more problematic!

PR #1213, opened 2023-05-31
Replace deprecated NIP-08 indexed mentions with NIP-27 nostr: prefixed URI text note references on writes

tyiu:
Changelog-Fixed: Replace deprecated NIP-08 indexed mentions with NIP-27 nostr: prefixed URI text note references on writes

https://github.com/damus-io/damus/assets/963907/3325c5e9-986b-423c-af4b-69bfa127c721

### DIFF
--- a/damus/Models/Mentions.swift
+++ b/damus/Models/Mentions.swift
@@ -10,7 +10,7 @@ import Foundation
 enum MentionType {
     case pubkey
     case event
-    
+
     var ref: String {
         switch self {
         case .pubkey:
@@ -495,14 +495,17 @@ func make_post_tags(post_blocks: [PostBlock], tags: [[String]], silent_mentions:
                 continue
             }
             
-            if let ind = find_tag_ref(type: ref.key, id: ref.ref_id, tags: tags) {
-                let mention = Mention(index: ind, type: mention_type, ref: ref)
+            if find_tag_ref(type: ref.key, id: ref.ref_id, tags: tags) != nil {
+                // Mention index is nil because indexed mentions from NIP-08 is deprecated.
+                // It has been replaced with NIP-27 text note references with nostr: prefixed URIs.
+                let mention = Mention(index: nil, type: mention_type, ref: ref)
                 let block = Block.mention(mention)
                 blocks.append(block)
             } else {
-                let ind = new_tags.count
                 new_tags.append(refid_to_tag(ref))
-                let mention = Mention(index: ind, type: mention_type, ref: ref)
+                // Mention index is nil because indexed mentions from NIP-08 is deprecated.
+                // It has been replaced with NIP-27 text note references with nostr: prefixed URIs.
+                let mention = Mention(index: nil, type: mention_type, ref: ref)
                 let block = Block.mention(mention)
                 blocks.append(block)
             }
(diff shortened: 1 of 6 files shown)


damusTests/ReplyTests.swift
```diff
@@ -397,7 +397,8 @@ class ReplyTests: XCTestCase {
     
     func testParsePostUriPubkeyReference() throws {
         let id = "6fec2ee6cfff779fe8560976b3d9df782b74577f0caefa7a77c0ed4c3749b5de"
-        let parsed = parse_post_blocks(content: "this is a nostr:p:\(id) event mention")
```
tyiu, line 400: @jb55 Was there a reason why we supported `nostr:p:hex` and `nostr:e:hex` URIs? It doesn't seem to be to spec according to NIP-21.
https://github.com/nostr-protocol/nips/blob/master/21.md

jb55: there was no spec when I wrote that

damus/Models/Mentions.swift
```diff
@@ -10,13 +10,16 @@ import Foundation
...
     
-    var ref: String {
+    var ref: String? {
         switch self {
         case .pubkey:
             return "p"
         case .event:
             return "e"
+        case .uri:
+            return nil
```
jb55, line 22: I don't think this is the right way to do it. we should be choosing pubkey or event depending on what's after the `nostr:`

jb55 (changes requested):

damus/Models/Mentions.swift
```diff
@@ -495,14 +498,17 @@ func make_post_tags(post_blocks: [PostBlock], tags: [[String]], silent_mentions:
                 continue
             }
             
-            if let ind = find_tag_ref(type: ref.key, id: ref.ref_id, tags: tags) {
-                let mention = Mention(index: ind, type: mention_type, ref: ref)
+            if find_tag_ref(type: ref.key, id: ref.ref_id, tags: tags) != nil {
+                // Mention index is nil because positional mentions from NIP-07 is deprecated.
```
tyiu, line 502: Typo: comment should say NIP-08, not NIP-07. For this comment and same comment below.

damus/Models/Mentions.swift
tyiu: Fixed in 120d6bd049bda8ba6c363ef9391f88518375df3b

damus/Models/Mentions.swift
tyiu: Fixed in 120d6bd049bda8ba6c363ef9391f88518375df3b

damus/Models/Post.swift
```diff
@@ -21,7 +21,6 @@ struct NostrPost {
     }
 }
 
-// TODO: parse nostr:{e,p}:pubkey uris as well
```
tyiu, line 24: Removed as it is not to spec.

tyiu:
Squashed 120d6bd049bda8ba6c363ef9391f88518375df3b into the original commit so that this branch has only one commit now.

tyiu:
@jb55 Thanks for the review. I addressed your feedback. Please take another look.

jb55:
nice, lgtm

tyiu:
Amended my commit to delete the `parse_nostr_ref_uri_type` function now that it's not being used anywhere.

PR #1222, opened 2023-06-01
added purplepag.es relay, and fixed mistype of purplepag.es relay add…

alltheseas:
added purplepag.es relay for profile discoverability, and fixed mistype of purplepag.es relay


tyiu (approved):

jb55:
Now that we have ephemeral relays (relay types), I would want to mark this relay as special in the code so we don't try to use it for all queries.

PR #1246, opened 2023-06-04
NostrScript

jb55:
NostrScript is a WebAssembly implementation that interacts with Damus. It enables dynamic scripting that can be used to power custom list views, enabling pluggable algorithms.

The web has JavaScript, Damus has NostrScript. NostrScripts can be written in any language that compiles to WASM.

This commit adds a WASM interpreter I've written as a mostly-single C file for portability and embeddability. In the future we could JIT-compile these for optimal performance if NostrScripts get large and complicated. For now an interpreter is simple enough for algorithm list view plugins.

Changelog-Added: Add initial NostrScript implementation


jb55:
This is probably against apple's ToS. I can make it so that NostrScript
is very limited in its functionality, but I still feel like apple would
be assholes and ban me just because we have wasm support.

How are we supposed to innovate on this platform? You could not build a
new protocol browser in 2023. I might need to wrap this around the
appstore flag and release in the freedom version only.

PR #1252, opened 2023-06-05
NEW Inter-font

BenGWeeks:
I have now applied the new Inter-font. For now, I haven't tried to scale the caption1, body, etc. as before which I think was causing issues.

CleanShot 2023-06-06 at 00 17 18

Seems to be OK with German (is this the right check @tyiu ?):

CleanShot 2023-06-06 at 00 21 40

PR is now rebased on the latest Master.


alltheseas:
I remember previously there were comments where Japanese was difficult to read in a new font. Have we tested with native 🇯🇵 speakers?

BenGWeeks:
> I remember previously there were comments where Japanese was difficult to read in a new font. Have we tested with native 🇯🇵 speakers?

CleanShot 2023-06-06 at 00 41 47

vs:

CleanShot 2023-06-06 at 00 44 15

NB Will be amending bold / exact sizes in re-branding.

tyiu:
iPhone SE is the phone to test against for small form factor.

BenGWeeks:
> iPhone SE is the phone to test against for small form factor.

I don't think the left navigation is very accessible as the font is fixed in size.

PR #1253, opened 2023-06-06
Translations

tyiu:


damus/Views/CreateAccountView.swift
```diff
@@ -167,7 +167,7 @@ func FormLabel(_ title: String, optional: Bool = false) -> some View {
         Text(title)
                 .bold()
         if optional {
-            Text("- optional", comment: "Label indicating that a form input is optional.")
```
tyiu, line 170: The `-` makes it a bit harder for translators to translate to other languages, and it isn't really necessary in the UI.

damus/Views/Profile/EditMetadataView.swift
```diff
@@ -103,13 +103,15 @@ struct EditMetadataView: View {
...
-                    TextField("Satoshi Nakamoto", text: $display_name)
+                    let display_name_placeholder = "Satoshi Nakamoto"
+                    TextField(display_name_placeholder, text: $display_name)
                         .autocorrectionDisabled(true)
                         .textInputAutocapitalization(.never)
                 }
                 
                 Section(NSLocalizedString("Username", comment: "Label for Username section of user profile form.")) {
-                    TextField("satoshi", text: $name)
+                    let username_placeholder = "satoshi"
```
tyiu, line 113: Extracting it out like this so that it doesn't get flagged as something that needs to be translated.

damus/Views/SetupView.swift
```diff
@@ -34,17 +34,12 @@ struct SetupView: View {
                         .shadow(color: DamusColors.purple, radius: 2)
                         .frame(width: 56, height: 56, alignment: .center)
                         .padding(.top, 20.0)
-                    
-                    HStack {
-                        Text("Welcome to", comment: "Welcome text shown on the first screen when user is not logged in.")
```
tyiu, line 39: The issue is that we can't split a sentence up into two like this. It makes it impossible for translators to translate, especially in languages that aren't grammatically the same as English. The compiler also doesn't like me embedding a gradient Text component inside a non-gradient Text component, for whatever reason.

I checked with Rob on the design, and he voted for applying the gradient on the entire string.

damus/en-US.xcloc/contents.json
```diff
@@ -3,10 +3,10 @@
   "project" : "damus.xcodeproj",
   "targetLocale" : "en-US",
   "toolInfo" : {
-    "toolBuildNumber" : "14E222b",
+    "toolBuildNumber" : "14E300c",
     "toolID" : "com.apple.dt.xcode",
     "toolName" : "Xcode",
-    "toolVersion" : "14.3"
+    "toolVersion" : "14.3.1"
```
tyiu, line 9: I upgraded my Xcode.

PR #1263, opened 2023-06-08
add relay logging

bryanmontz:
This PR adds relay logging to the RelayDetailView.

Features:
* list of events, most recent at the top
* shows changes to RelayConnection's state
* shows changes to network state
* RelayDetailView updates live as lines are added to the log
* logs persisted to disk for later investigation
* line limit prevents unbounded file growth

Here's how it looks when resuming after having the app in the background for a while and the relays need to reconnect:  

relay-log


jb55:
I've been hesistant to merge this because I feel like most users would not need it. Perhaps this could be gated behind a developer switch?

Issue #1238, opened 2023-06-03
Videos sounds can play over each other


punterwantsawhalepass:
If you start a video and enable the sound, you can then go to another video and start it. You will then have two videos sound playing at the same time, the first one and the second one you are now watching.

Cool feature :) but may be better for app to cut off first video sound from streaming when the second one starts.

alltheseas:
From @dmnyc 

> What happens:
In the new video player, if the sound is on and the video scrolls out of frame in timeline view, the sound continues playing on a loop.

> What I think should happen:
Video should pause or mute if it is no longer visible on screen.

Example note:
note16dvkw4tur4t7qljvnkq4knuey74l525f70d3hmxzqnw3mqhu798q7kt6dr

alltheseas:
From @BenGWeeks 

https://github.com/damus-io/damus/assets/64376233/3811777c-15e6-4890-835d-6107da78d8b2

alltheseas:
https://damus.io/note1jdjdkh2xphnz0lajpxvz6jx4xq3ffeuhflfguz4wkne6gt4e6qcss8uycq

gladiusKatana:
By the way, funny story, apparently one of the big centralized social media apps had this bug as well recently😅 (the person who told me that thought it "might have been facebook?" and that it was on the web...I'll mention it here if i find out which!)

alltheseas:
This seems fixed 1.5 (8)

PR #1266, opened 2023-06-08
[UX] Fix Issue 1238: Videos sounds can play over each other

gladiusKatana:
Fix issue #1238 - make videos automatically pause after scrolling away from a post containing a video

Fix
-Auto-pause triggered in an .onChange that compares the y-coordinate of the center of a video with the screen visible area
-Tolerance can be set according to design preferences (ie. how far user needs to scroll before it pauses) 

Demo

https://github.com/damus-io/damus/assets/47217795/2cb126db-9b55-4c78-975f-3d692a0b0a9d

*handy account that posts quite a few videos (i just spun up a new account & followed only this one) :
npub1ztdt0nvu8324tg40g4ts8d6englps8w6574zyq2gkdvcrx364y0qd9x9za

### DIFF
--- a/damus/Views/Video/DamusVideoPlayer.swift
+++ b/damus/Views/Video/DamusVideoPlayer.swift
@@ -43,25 +43,34 @@ struct DamusVideoPlayer: View {
     }
     
     var body: some View {
-        ZStack(alignment: .bottomTrailing) {
-            VideoPlayer(url: url, model: model)
-                .onAppear {
-                    model.start()
+        GeometryReader { geo in
+            let localFrame = geo.frame(in: .local)
+            let localCenter = CGPoint(x: localFrame.midX, y: localFrame.midY)
+            let globalCenter = geo.frame(in: .global).origin.applying(.init(translationX: localCenter.x, y: localCenter.y))
+            let centerY = globalCenter.y
+
+            ZStack(alignment: .bottomTrailing) {
+                VideoPlayer(url: url, model: model)
+                if model.has_audio == true {
+                    MuteIcon
+                        .zIndex(11.0)
+                        .onTapGesture {
+                            self.model.muted = !self.model.muted
+                        }
                 }
-            
-            if model.has_audio == true {
-                MuteIcon
-                    .zIndex(11.0)
-                    .onTapGesture {
-                        self.model.muted = !self.model.muted
-                    }
             }
-        }
-        .onChange(of: model.size) { size in
-            guard let size else {
-                return
+            .onChange(of: model.size) { size in
+                guard let size else {
+                    return
+                }
+                video_size = size
             }
-            video_size = size
+            .onChange(of: centerY) { _ in
+                let screenHeight = UIScreen.main.bounds.height
+                let screenMidY = screenHeight / 2
+                let tol = 0.20 * screenHeight /// tolerance - can vary  to taste ie.,  %  of screen height of a centered box in which video plays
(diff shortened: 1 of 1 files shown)


bryanmontz:
A fix that is based on introducing a slight delay usually indicates that the problem is not fully understood. It will make future understanding and debugging more challenging.

I think the real issue is that video metadata and video playback state are cached together. They should be separated so that the metadata can be cached (has_audio and size), but the playback state is new with any view created for playback.

gladiusKatana:
> A fix that is based on introducing a slight delay usually indicates that the problem is not fully understood. It will make future understanding and debugging more challenging.
> 
> I think the real issue is that video metadata and video playback state are cached together. They should be separated so that the metadata can be cached (has_audio and size), but the playback state is new with any view created for playback.

RE: delay - my bad - fixed temporary commit (changes copied below)

RE. caching: yes good idea i think. Looking into how else playback state (and/or the video asset itself) could be managed. Some findings:
  -playback state is touched in updateUIView & makeUIView as expected… -…BUT current model.start() call in .onAppear in master _isn’t needed_ for current functionality (videos play, looped, when in-view & stop when view is deallocated — as you said: “new with any view created” 
-Strangely, in model: methods start() & stop() aren’t called anywhere (else); and the 4 methods below them aren’t called _anywhere_; And model.autoReplay will never be false (obviating a conditional in makeUIView)

Even notwithstanding any playback logic cleanup … not sure yet _where_ besides updateUIView playback state could be controlled dynamically…  IF video (asset itself?) were deallocated “immediately” on onDisappear (assuming any content offset compensation we might need or detecting onDisappear from some small invisible overlay view etc.), couldn’t we in principle control playback state regardless of if/how video/model is cached?   (Maybe i’ll ping you in Chat - i know you know Damus’s data layer / caching deeply, maybe i’m missing something key)

Changes

```swift
.onAppear {
                    model.start()
                }
                .onDisappear() {
                    model.stop()
                }
```

gladiusKatana:
PS. another earlier thought: since onDisappear also currently gets called with latency (probably due to content offsets at least in part), I wonder whether we _in principle_ could control playback-state ‘instantly’ (even if in addition to video/playback caching changes) by, say, updating model via a small invisible overlay view’s .onDisappear (basically tracking when video moves away from user’s immediate field of view)

gladiusKatana:
ok - found a much more robust and fine-tuneable solution - will ping you back @bryanmontz 
and updated comment above (will record new demo tomorrow)

bryanmontz:
This is definitely getting close, feels much better than before. But I don't understand why you need the two rectangles. It seems like you could just make the play/pause decision based on the `globalFrame`s `minY` and `height`. Is it near the top of the screen or near the bottom of the screen? Also why do you need to use `localFrame.midX` for the `videoRect`? Why isn't the x value of that just 0?

gladiusKatana:
> on the `globalFrame`s `minY`

Yeah good thought re. using globalFrame's minY; what I've found though is, like w/ .onDisappear, an .onChange of the minY never gets called(!) even though _this_ minY should in theory change as you scroll around. But for some reason only the 2nd of these 2 onChanges below _does_ get called ... it's as if SwiftUI "likes intersections better than frame parameters alone" in terms of tracking changes. Seems kind of arbitrary...

But you're right i think, we're close (update you here/in chat when fix is found)

jb55:
This seems much better now. Merged in df20b67fc12d10bec362e79f2ef7f4205f532871

PR #1274, opened 2023-06-11
Add copy button to relay id and lud16 in wallet.

### LINKED ISSUE #1273: Feature Request - Add copy to clipboard button to NWC page
Damus supports NWC, allowing you to connect certain lightning-enabled wallets (such as Alby) to connect and take advantage of one-tap-zaps.

IMG_6249

The connection process is straight forward and can be done entirely from your iOS device. However, in a situation where the wallet connected to Damus runs out of sats or you are needing to refill it for a larger transaction, the process is a bit lengthy to process on the go. Attached is the current process if only using mobile and assuming one is not using a web browser.

https://github.com/damus-io/damus/assets/111670064/48f06a38-b64b-4f7a-b87a-0be0e4e10e09

I propose adding a copy button, similar to the option to copy your npub on the profile page. This would allow a user to quickly open the wallet page, copy their lightning address, paste it into a lightning wallet app (like Wallet of Satoshi) and refill their wallet that is connected to Damus.

suhailsaqan:
Fixes #1273

Screenshot 2023-06-10 at 10 16 16 PM

### DIFF
--- a/damus/Views/Wallet/WalletView.swift
+++ b/damus/Views/Wallet/WalletView.swift
@@ -18,16 +18,36 @@ struct WalletView: View {
         self._settings = ObservedObject(wrappedValue: damus_state.settings)
     }
     
+    func copyText(_ text: String) {
+        UIPasteboard.general.string = text
+    }
+    
     func MainWalletView(nwc: WalletConnectURL) -> some View {
         VStack {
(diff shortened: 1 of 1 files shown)


jb55:
I'd like to implement the wallet design before changing anything here

PR #1277, opened 2023-06-11
fix: Show DM error rather than editing event content

realprogrammersusevim:
When a DM event cannot be decrypted Damus currently edits the content of the DM to say `*failed to decrypt content*`. This is problematic for two reasons:

1. We shouldn't edit the content of an event to display an error to avoid actually dealing with that error
2. Damus no longer renders Markdown for events so `*failed to decrypt content*` is no longer correctly italicized.

To fix this I added a `decryptable` variable to the NostrEvent type that defaults to true but is set to false when a DM cannot be decrypted. This value is only read and used when displaying DMs and allows Damus to show a styled text event rather than broken Markdown plaintext so the user has more of a hint that the text is coming from Damus rather than a strange message by another user.

I also added the `too_big` variable into the `should_show_event` since they were already the same thing.

### DIFF
--- a/damus/Nostr/NostrEvent.swift
+++ b/damus/Nostr/NostrEvent.swift
@@ -161,7 +161,7 @@ class NostrEvent: Codable, Identifiable, CustomStringConvertible, Equatable, Has
                 return decrypted_content
             } else {
                 self.decryptable = false
-                return "Error: failed to decrypt content"
+                return ""
             }
         }
         
--- a/damus/Nostr/NostrEvent.swift
+++ b/damus/Nostr/NostrEvent.swift
@@ -66,6 +66,7 @@ class NostrEvent: Codable, Identifiable, CustomStringConvertible, Equatable, Has
     let created_at: Int64
     let kind: Int
     let content: String
+    var decryptable: Bool = true
     
     var is_textlike: Bool {
         return kind == 1 || kind == 42
@@ -156,7 +157,12 @@ class NostrEvent: Codable, Identifiable, CustomStringConvertible, Equatable, Has
 
     func get_content(_ privkey: String?) -> String {
         if known_kind == .dm {
-            return decrypted(privkey: privkey) ?? "*failed to decrypt content*"
+            if let decrypted_content = decrypted(privkey: privkey) {
+                return decrypted_content
+            } else {
+                self.decryptable = false
+                return "Error: failed to decrypt content"
+            }
         }
         
         return content
--- a/damus/Views/NoteContentView.swift
+++ b/damus/Views/NoteContentView.swift
@@ -106,7 +106,7 @@ struct NoteContentView: View {
     var MainContent: some View {
         VStack(alignment: .leading) {
             if !self.event.decryptable {
-                Text("Error: failed to decrypt content").italic().bold()
+                Text("Error: failed to decrypt content", comment: "Text error shown when Damus cannot decipher and understand a Direct message").italic().bold()
             } else {
                 if size == .selected {
                     if with_padding {
--- a/damus/Views/NoteContentView.swift
+++ b/damus/Views/NoteContentView.swift
@@ -105,55 +105,58 @@ struct NoteContentView: View {
     
     var MainContent: some View {
         VStack(alignment: .leading) {
-            if size == .selected {
-                if with_padding {
-                    SelectableText(attributedString: artifacts.content.attributed, size: self.size)
-                        .padding(.horizontal)
-                } else {
-                    SelectableText(attributedString: artifacts.content.attributed, size: self.size)
-                }
+            if !self.event.decryptable {
+                Text("Error: failed to decrypt content").italic().bold()
             } else {
-                if with_padding {
-                    truncatedText
-                        .padding(.horizontal)
+                if size == .selected {
+                    if with_padding {
+                        SelectableText(attributedString: artifacts.content.attributed, size: self.size)
+                    } else {
+                        SelectableText(attributedString: artifacts.content.attributed, size: self.size)
+                    }
                 } else {
-                    truncatedText
+                    if with_padding {
+                        truncatedText
+                            .padding(.horizontal)
+                    } else {
+                        truncatedText
+                    }
                 }
-            }
-
-            if !options.contains(.no_translate) && (size == .selected || damus_state.settings.auto_translate) {
-                if with_padding {
-                    translateView
-                        .padding(.horizontal)
-                } else {
-                    translateView
+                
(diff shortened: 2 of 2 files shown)


alltheseas:
Can you show a before/after side-by-side video/animation for reviewers benefit 🙏

jb55:
On Sun, Jun 11, 2023 at 01:09:09PM -0500, Jonathan wrote:
>The variable too_big in NostrEvent was only refered to by should_show
>event so I combined the two into should_show_event.
>---
> damus/Nostr/NostrEvent.swift | 10 +++++-----
> 1 file changed, 5 insertions(+), 5 deletions(-)
>
>diff --git a/damus/Nostr/NostrEvent.swift b/damus/Nostr/NostrEvent.swift
>index 2879ddef..d56a0f0f 100644
>--- a/damus/Nostr/NostrEvent.swift
>+++ b/damus/Nostr/NostrEvent.swift
>@@ -71,12 +71,12 @@ class NostrEvent: Codable, Identifiable, CustomStringConvertible, Equatable, Has
>         return kind == 1 || kind == 42
>     }
>
>-    var too_big: Bool {
>-        return self.content.utf8.count > 16000
>-    }
>-
>     var should_show_event: Bool {
>-        return !too_big
>+        if self.content.utf8.count > 1600 {

This changes the behavior from 16000 to 1600 which is pretty low

This also changes the behavior in another way: this will always return
false with small contents which is also incorrect. Maybe best to just
drop this patch from your PR.

>+            return true
>+        } else {
>+            return false
>+        }
>     }
>
>     var is_valid_id: Bool {
>-- 
>2.39.2 (Apple Git-143)

realprogrammersusevim:
That commit has been reverted. Anything else need to be cleaned up?

jb55:
On Sun, Jun 25, 2023 at 04:16:50PM -0700, Jonathan wrote:
>That commit has been reverted. Anything else need to be cleaned up?

can you rebase, dropping the first commit and its revert commit ?

realprogrammersusevim:
Done! The commit history is clean now.

jb55:
On Sun, Jun 11, 2023 at 01:11:59PM -0500, Jonathan wrote:
>When a DM could not be decrypted the `get_content` function would edit
>the content of the event. Now it changes the `decryptable` variable to
>false which allows for styled text when displaying the undecryptable
>event.
>
>Closes: https://github.com/damus-io/damus/pull/1277
>---
> damus/Nostr/NostrEvent.swift      |  8 ++-
> damus/Views/NoteContentView.swift | 83 ++++++++++++++++---------------
> 2 files changed, 50 insertions(+), 41 deletions(-)
>
>diff --git a/damus/Nostr/NostrEvent.swift b/damus/Nostr/NostrEvent.swift
>index 2879ddef1..6fc480b88 100644
>--- a/damus/Nostr/NostrEvent.swift
>+++ b/damus/Nostr/NostrEvent.swift
>@@ -66,6 +66,7 @@ class NostrEvent: Codable, Identifiable, CustomStringConvertible, Equatable, Has
>     let created_at: Int64
>     let kind: Int
>     let content: String
>+    var decryptable: Bool = true
>
>     var is_textlike: Bool {
>         return kind == 1 || kind == 42
>@@ -156,7 +157,12 @@ class NostrEvent: Codable, Identifiable, CustomStringConvertible, Equatable, Has
>
>     func get_content(_ privkey: String?) -> String {
>         if known_kind == .dm {
>-            return decrypted(privkey: privkey) ?? "*failed to decrypt content*"
>+            if let decrypted_content = decrypted(privkey: privkey) {
>+                return decrypted_content
>+            } else {
>+                self.decryptable = false
>+                return "Error: failed to decrypt content"
>+            }
>         }
>
>         return content
>diff --git a/damus/Views/NoteContentView.swift b/damus/Views/NoteContentView.swift
>index 5ca6942df..1c2bb54ef 100644
>--- a/damus/Views/NoteContentView.swift
>+++ b/damus/Views/NoteContentView.swift
>@@ -105,57 +105,60 @@ struct NoteContentView: View {
>
>     var MainContent: some View {
>         VStack(alignment: .leading) {
>-            if size == .selected {
>-                if with_padding {
>-                    SelectableText(attributedString: artifacts.content.attributed, size: self.size)
>-                        .padding(.horizontal)
>-                } else {
>-                    SelectableText(attributedString: artifacts.content.attributed, size: self.size)
>-                }
>+            if !self.event.decryptable {

I think this can be greatly simplified: instead of adding another nested
if statement, we can do something like:

     var ArtifactContent: some View {
         Group {
             switch self.note_artifacts {
             case .longform(let md):
                 Markdown(md.markdown)
                     .padding([.leading, .trailing, .top])
             case .separated(let separated):
		if !event.decryptable {
			Text("Error: failed to decrypt content").italic().bold()
		} else {
			MainContent(artifacts: separated)
		}
             }
         }
         .fixedSize(horizontal: false, vertical: true)
     }

jb55:
On Sun, Jun 11, 2023 at 01:37:05PM -0500, Jonathan wrote:
>Added a comment to the error for the translators. Also changed the
>returned string from an undecryptable NostrEvent to ""
>
>Closes: https://github.com/damus-io/damus/pull/1277
>---
> damus/Nostr/NostrEvent.swift      | 2 +-
> damus/Views/NoteContentView.swift | 2 +-
> 2 files changed, 2 insertions(+), 2 deletions(-)
>
>diff --git a/damus/Nostr/NostrEvent.swift b/damus/Nostr/NostrEvent.swift
>index 6fc480b88..35d40f550 100644
>--- a/damus/Nostr/NostrEvent.swift
>+++ b/damus/Nostr/NostrEvent.swift
>@@ -161,7 +161,7 @@ class NostrEvent: Codable, Identifiable, CustomStringConvertible, Equatable, Has
>                 return decrypted_content
>             } else {
>                 self.decryptable = false
>-                return "Error: failed to decrypt content"
>+                return ""
>             }
>         }
>
>diff --git a/damus/Views/NoteContentView.swift b/damus/Views/NoteContentView.swift
>index 1c2bb54ef..b27dab29d 100644
>--- a/damus/Views/NoteContentView.swift
>+++ b/damus/Views/NoteContentView.swift
>@@ -106,7 +106,7 @@ struct NoteContentView: View {
>     var MainContent: some View {
>         VStack(alignment: .leading) {
>             if !self.event.decryptable {
>-                Text("Error: failed to decrypt content").italic().bold()
>+                Text("Error: failed to decrypt content", comment: "Text error shown when Damus cannot decipher and understand a Direct message").italic().bold()

Let's just combine these commits into one.

PR #1296, opened 2023-06-17
simplify and fix issues with RelayStatus

bryanmontz:
While attempting to debug the recent relay connection issues, I noticed some problems and improvement opportunities with `RelayStatus`.

This PR addresses these issues:
1. The console message "No image named 'network' found in asset catalog..." prints for every occurrence of `RelayStatus`, presumably because that image was removed with the iconography update.
2. The property `connecting` never gets set to false after being set to true, so you have to leave the view and come back for it to update. Otherwise it will always appear to be connecting even after it has reconnected.
3. The view uses a timer for polling the connection status every 2 seconds. This PR updates that to make `RelayConnection` observable.
4. `RelayStatus` was being passed the entire `RelayPool` when all it needs is a `RelayConnection` to track.

Lastly, in a separate commit, I changed the name of `RelayStatus` to `RelayStatusView`. IMO it makes code easier to read and maintain when view names end with the word "view".


jb55:
thanks! merged

PR #1298, opened 2023-06-17
after loading the user's relays from their contact event, connect to new relays

bryanmontz:


bryanmontz:
`RelayPool.connect(to:)` was being called before all the relays were added to the pool. The hard-coded bootstrap relays were always in place in time for that `connect`, but then when the app requests the user's contact event so we can load their relays, its too late and those relays never get `connect` called on them. Subsequent app launches work as expected.

tyiu (approved):

jb55:
On Sat, Jun 17, 2023 at 06:58:22AM -0700, Bryan Montz wrote:
>`RelayPool.connect(to:)` was being called before all the relays were added to the pool. The hard-coded bootstrap relays were always in place in time for that `connect`, but then when the app requests the user's contact event so we can load their relays, its too late and those relays never get `connect` called on them. Subsequent app launches work as expected.

makes sense!

PR #1311, opened 2023-06-22
Changelog-Added: loaded a handful of new icons and separated the curr…

robagreda:
Changelog-Added: loaded a handful of new icons and separated the current ones into regular and regular.fill to have for each icon two versions, one is outlined and the other is the filled version. Updated Strike App icon and added new Hashtag icons.

@BenGWeeks could you please check if everything works fine with your current icons that you've added compared to these ones that some of them are replaced.


jb55:
there are a bunch of duplicate files in this PR. I can look into writing a script to conslidate them but I'm not sure when I'll get around to this.

PR #1321, opened 2023-06-25
fix crash related to VideoPlayer and CMTime

bryanmontz:
Changelog-Fixed: fix crash related to VideoPlayer and CMTime

This PR fixes a crash related to VideoPlayer and CMTime.

The `VideoPlayerModel`s `time` property was initialized with an invalid `CMTime` (e.g. `CMTime()`), and then at some point the `VideoPlayer` tried to seek to that time.

The fix is to make that `CMTime` optional and to check if the time is valid using Apple's framework's built-in checks before seeking.

Another minor issue was that when seeking the `toleranceBefore` and `toleranceAfter` values were set to the same `CMTime` as the time being seeked to. This doesn't add much and isn't necessary, as we can call `AVPlayer.seek(to:completion:)` instead and not bother with the tolerances.

### DIFF
--- a/damus/Views/Video/VideoPlayer.swift
+++ b/damus/Views/Video/VideoPlayer.swift
@@ -48,7 +48,8 @@ public class VideoPlayerModel: ObservableObject {
     @Published var has_audio: Bool? = nil
     @Published var contentMode: UIView.ContentMode = .scaleAspectFill
     
-    var time: CMTime = CMTime()
+    fileprivate var time: CMTime?
+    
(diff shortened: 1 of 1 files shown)


tyiu (approved):

jb55:
Thank you, I was trying to figure this one out and I didn't know what
was happening... there was no information in the exception?

bryanmontz:
When I reproduced it in the simulator, it said "Seeking is not possible to time {INVALID}".  
damus_—_damusApp_swift

PR #1328, opened 2023-06-26
Issue 1286

thefloofe22:
This is a fix for issue 1286. Let me know if anything doesn't look right

### DIFF
--- a/damus-c/nostr_bech32.c
+++ b/damus-c/nostr_bech32.c
@@ -9,6 +9,7 @@
 #include <stdlib.h>
 #include "cursor.h"
 #include "bech32.h"
+#include <stdio.h>
 
 #define MAX_TLVS 16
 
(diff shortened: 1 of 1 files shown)


tyiu:
Would you mind explaining the problem and the fix? It's not clear from looking at the PR title, PR description, or code. It would be great if you could update them.

thefloofe22:
Yes, the issue is that there is a generic error message when media that is greater than 10 mb is uploaded to nostr.build. This results in a user experience that is confusing is they don't understand that it is due to the fact that media uploaded is too large. Hence, the idea was the add a specific error message. I believe I isolated the part of the code that does this correctly, and i used the fprintf function to stderr to return the error. Kindly let me know if any of my thinking or code is incorrect, and I will fix it accordingly.

jb55:
This code modified has nothing to do with the issue you are referring to.

PR #1332, opened 2023-06-27
Fix missing profile zap notification text

tyiu:
Changelog-Fixed: Fix missing profile zap notification text


damus/Views/Notifications/EventGroupView.swift
```diff
@@ -252,6 +252,7 @@ struct EventGroupView: View {
                     }
                     .buttonStyle(.plain)
                 } else {
+                    GroupDescription(unique_pubkeys)
```
tyiu, line 255: This line was accidentally reverted in 66db4c521576dcfcb5470d561d06ec4a88716adb

PR #1353, opened 2023-07-08
Minor update to ConfigView.swift to fix key and search icon

cr0bar:
Replaced icon names with names existing in assets to avoid them being left blank.


jb55:
awesome, thanks!

Issue #1352, opened 2023-07-07
Malformed user link when prefixed by invisible unicode characters


adebroise:
Seems user reference (like @ Will ) don’t render properly when immediately preceded by another character. For example, I had one yesterday where it output the nostr key, looking at a dump of the note, it was prefixed with a zero width Unicode character (U+200B). Seems to work on other clients like Snort.

![image](https://github.com/damus-io/damus/assets/69913674/3654aa95-b614-42f6-98fa-e86f6fdf07a6)

![image](https://github.com/damus-io/damus/assets/69913674/9aa4fca2-bc2d-4e54-8344-ceed80405f81)

cr0bar:
Similar issue if the user reference is at the start of a note. Here the bottom one is where the @ is the first character, and the top is where a space is added before the @.
![IMG_8591](https://github.com/damus-io/damus/assets/11259194/e29b8db9-47f2-4dc6-8664-b14b2d1f26a0)

alltheseas:
Thanks gents.

Related to URL preceded by emoji issue https://github.com/damus-io/damus/issues/1268

cr0bar:
Follow up to the screenshot above. It seems when you post a note using reply it’s using the hex version of the public key and when you manually type a note it’s using a bech32 version of the public key.

Reply (2nd example in screenshot):

{"content":"**nostr:32e1827635450ebb3c5a7d12c1f8e7b2b514439ac10a67eef3d9fd9c5c68e245** enjoying having the full fat #damus experience back again 😁","tags":[["t","damus"]],"id":"5ec9c226a5a77595257f79e620ab9b4f45d1cb89062b28a51982fc10fb6f6cb1","kind":1,"created_at":1688749391,"sig":"b5e46c45ed0d4fa0911a2adc32daa39d36ba705d704b45b928ed9d0698a45f6d26d50579be2b2845274b960cd77765b5893bf8462f029e20f187714027e4e311","pubkey":"f16ac67afff1340501cf27b7ebe9e3086fb62e97f044ba407e38b5d8425a73f2"}

Note (first example in screenshot):

[...]

PR #1355, opened 2023-07-08
Update to PostView.swift to fix broken nostr link

cr0bar:
Fix for second part of issue #1352 where if you submit a reply from the + on a profile, it uses the hex nostr url rather than the bech32 version. When typing the @ manually it uses the bech32 so updated to mirror this.


jb55:
thank you!

PR #1358, opened 2023-07-09
Update CustomizeZapView.swift; zap ~~user~~

olheadinternet:
Change the text "zap user" back to "zap". If "nozap" mode is true, this page doesn't show at all now. with "nozap" mode false it does functions as a note zap, so the text now reflects that.


olheadinternet:
actually... i think this effects both customize sheets... brb.

Issue #1362, opened 2023-07-11
Hashtags don’t work from search result with replies


cr0bar:
**Issue:**

There seems to be a bug with hashtags where they’re displayed alongside a reply on the back of a search. Bug exists in current AppStore version.

**Reproduce:**

- Go into Search and search for #damus
- Select a note which has replies
- Attempt to tap the hashtag in the initial note

The hashtag will not return to the search screen for the selected hashtag.

cr0bar:
Working when in this view:

![image](https://github.com/damus-io/damus/assets/11259194/4b25f1bb-ed37-40f4-aa16-93b800078661)

But not in this view:

![image](https://github.com/damus-io/damus/assets/11259194/29ff1ae6-2147-49e2-b0c9-a9c642f967a6)

cr0bar:
If you switch rotation from portrait to horizontal and back, they work. Seems to only affect SelectableText not truncateText.

alltheseas:
I cannot seem to recreate on 1.7 (11). @cr0bar is this still an issue for you?

alltheseas:
> **Issue:**
> 
> There seems to be a bug with hashtags where they’re displayed alongside a reply on the back of a search. Bug exists in current AppStore version.
> 
> **Reproduce:**
> 
> * Go into Search and search for #damus
> * Select a note which has replies
> * Attempt to tap the hashtag in the initial note
> 
> The hashtag will not return to the search screen for the selected hashtag.

first time: following this brings me back to the hashtag screen, but the content never loads, and the app freezes
second time and third time: app freezes in hashtag search view

1.17 (1299)

danieldaquino:
> first time: following this brings me back to the hashtag screen, but the content never loads, and the app freezes
second time and third time: app freezes in hashtag search view

Possibly related issue: https://github.com/damus-io/damus/issues/3795

PR #1366, opened 2023-07-12
Fix for broken hashtags on first load of a thread

cr0bar:
Fixes: Hashtags don’t work from search result with replies #1362


cr0bar:
Didn't have desired effect, will re-visit.

PR #1372, opened 2023-07-16
Include .m3u8 for video playback to enable streaming

fishcakeday:
This trivial change should enable HLS streaming and adaptable bit rate playback for video playback.


jb55:
On Sun, Jul 16, 2023 at 06:43:45AM -0700, Fishcake wrote:
>Although it starts m3u8 playback for VOD, the size and aspect ration is wrong and broken.

is there a note I could test this on?

Issue #104, opened 2022-12-22
Profile View pushed in the navigation infinitely


scoder1747:
![ezgif com-gif-maker](https://user-images.githubusercontent.com/120697811/209230174-44b4ceb5-3357-49e8-b215-f900400fc10e.gif)

jb55:
I have never experienced this. bizarre

jb55:
oh I see what you mean. yeah we can fix this

alltheseas:
I think this is fixed @scoder1747. Can you close if fixed?

scoder1747:
> I think this is fixed @scoder1747. Can you close if fixed?

This behavior still persists. The user shouldn't be navigated to the profile by tapping on the image/link if he is viewing the same profile.

cc @jb55

alltheseas:
@scoder1747 I don't understand how to recreate this issue. 

Could you please start from e.g. closing Damus app, opening Damus app, and recording your screen from there 🙏

alltheseas:
I see what you mean now @scoder1747

alltheseas:
Related: infinite hashtag rabbit hole

jb55:
Not fixed yet

jb55:
nm this is fixed, just not hashtags yet

PR #1390, opened 2023-07-26
FIX: Profile View pushed in the navigation infinitely #104

brett-doffing:
This fix adds a check, when creating `EventShell`s, to look at the last path of the NavigationCoordinator, in order to determine if the last view (current view) is a profile view, via its `Route`. If it is, and the public key of the `EventShell` is the same as the profile, then the pfp will not be wrapped in a nav link, preventing the ability to push the same profile infinitely.

### DIFF
--- a/damus/Views/Events/EventShell.swift
+++ b/damus/Views/Events/EventShell.swift
@@ -52,21 +52,16 @@ struct EventShell<Content: View>: View {
     }
 
     func Pfp(is_anon: Bool) -> some View {
-        // Prevent navigation from pfp if on profile
-        var preventNavLink = false
+        var profilePubKey: String?
         
         // Look at nav stack to determine if last (current) view is a profile, via its route,
         // then compare pubKeys
         if let route = state.nav.path.last {
             switch route {
-            case .ProfileByKey(let profilePubKey):
-                if profilePubKey == pubkey {
-                    preventNavLink = true
-                }
+            case .ProfileByKey(let profileKey):
+                profilePubKey = profileKey
             case .Profile(let profileModel, _):
-                if profileModel.pubkey == pubkey {
-                    preventNavLink = true
-                }
+                profilePubKey = profileModel.pubkey
             default:
                 break
             }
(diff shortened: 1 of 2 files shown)


jb55:
Can you squash your changes? It would make it easier for me to review as I do not use github when reviewing code (I rely on pulling patches for a PR from their server). Thanks!

jb55:
Hey! Thanks for the submission.

A few comments below:

On Wed, Jul 26, 2023 at 12:54:43PM -0500, doffing.brett wrote:
>Created flag, moved check, add `Route.Profile` check
>
>Change comparison!
>
>Cleanup check

needs this in the commit log:

Changelog-Fixed: Prevent nested profile navigation

>+        // then compare pubKeys
>+        if let route = state.nav.path.last {
>+            switch route {
>+            case .ProfileByKey(let profilePubKey):
>+                if profilePubKey == pubkey {
>+                    preventNavLink = true
>+                }
>+            case .Profile(let profileModel, _):
>+                if profileModel.pubkey == pubkey {
>+                    preventNavLink = true
>+                }
>+            default:
>+                break
>+            }
>+        }

You can use where pattern guards here:

if let route = state.nav.path.last {
     switch route {
     case .ProfileByKey(let profilePubKey) where profilePubKey == pubkey,
          .Profile(let profileModel, _) where profileModel.pubkey == pubkey:
         preventNavLink = true
     default:
         break
     }
}

brett-doffing:
I get "must be bound in every pattern" errors? Regardless, I can remove repetition and nesting.

jb55:
On Thu, Jul 27, 2023 at 12:06:36PM -0700, doffing.brett wrote:
>I get "must be bound in every pattern" errors? Regardless, I can remove repetition and nesting.

I could be wrong, I just did it off the top of my head. If you couldn't
get it working then don't worry about it.

brett-doffing:
I believe the enum cases simply need to have the same associated value when trying to do inline stuff like this. I had thought about replacing the two if statements with one because it seems a little cleaner, so I just implemented that alongside the commit message you suggested.

jb55:
The profile pic view is used in many places. We will need a
"NavProfilePicView" that does this logic all in one place.

jb55:
what do you think of this more general solution?

https://github.com/damus-io/damus/issues/1401

brett-doffing:
> what do you think of this more general solution?

I can't argue against it.

brett-doffing:
Closing in favor of more general solution at #1401

PR #1393, opened 2023-07-28
Initial nostrdb integration

jb55:
This is a rewrite of the codebase. There was no reasonable way to break this into smaller commits, so here it is.

- Switch to nostrdb's NdbNote type instead of NostrEvents
- Much improved typing, Pubkeys are now Pubkeys (raw data) instead of String

This prepares the codebase for the switch over to memory-mapped note data which will live inside lmdb.


jb55:
these commits are too big. since I'm seeing a few bugs on this branch I am going to start making the commits more granular so I can bisect issues

Issue #1150, opened 2023-05-19
Save image is broken (fix image permissions)


alltheseas:
_what happens_

When i long tap an image and hit save image, image does not save

_what i think should happen_

When i long tap an image and hit save image, image should save

brett-doffing:
Has this been fixed? I can't seem to reproduce.

brett-doffing:
Actually, I just turned off photos permissions and it did not save, for which I imagine the user should be prompted

brett-doffing:
I have implemented an alert if there is an error saving a photo, although I don't know if this is a bug outside of permissions, so I wouldn't exactly know what I should display for the alert message.

jb55:
On Fri, Jul 28, 2023 at 11:39:16AM -0700, doffing.brett wrote:
>I have implemented an alert if there is an error saving a photo, although I don't know if this is a bug outside of permissions, so I wouldn't exactly know what I should display for the alert message.

I'm guessing the permission was denied and we don't prompt it again?

alltheseas:
Sounds like a reasons explanation

alltheseas:
Closing as part of a backlog triage — we're narrowing focus to the most actionable issues to reduce noise for contributors.

If this is still relevant, please reopen with updated context or reproduction steps. Thanks for reporting!

PR #1403, opened 2023-07-31
Changelog-Fixed: Save Image Permissions

brett-doffing:
Currently, without permission, saving an image will fail silently and simply not save the image. This PR adds an `ImageSaver` class to handle permission failures when saving images, as well as display an alert upon failure.

[Fixes: #1150]

### DIFF
--- a/damus/Util/Images/ImageSaver.swift
+++ b/damus/Util/Images/ImageSaver.swift
@@ -1,0 +1,27 @@
+// ImageSaver.swift
+
+import UIKit
+
+enum ImageSaverError {
+    case permissions
+    case unknown
+}
+
+class ImageSaver: NSObject, ObservableObject {
+    var errorTitle: String = "Failed to save"
+    var errorMessage: String = "Unable to save the photo."
+    var errorType: ImageSaverError = .unknown
+    @Published var error = false
+
+    func writeToPhotoAlbum(image: UIImage) {
+        UIImageWriteToSavedPhotosAlbum(image, self, #selector(saveCompleted), nil)
+    }
+
+    @objc func saveCompleted(
+        _ image: UIImage,
+        didFinishSavingWithError error: Error?,
+        contextInfo: UnsafeRawPointer
+    ) {
+        self.error = error != nil
+    }
+}
--- a/damus/Views/Images/ImageContainerView.swift
+++ b/damus/Views/Images/ImageContainerView.swift
@@ -15,7 +15,8 @@ struct ImageContainerView: View {
     
     @State private var image: UIImage?
     @State private var showShareSheet = false
-    
+    @ObservedObject var imageSaver = ImageSaver()
+
     let disable_animation: Bool
     
     private struct ImageHandler: ImageModifier {
@@ -35,10 +36,23 @@ struct ImageContainerView: View {
             }
             .imageModifier(ImageHandler(handler: $image))
             .clipped()
-            .modifier(ImageContextMenuModifier(url: url, image: image, showShareSheet: $showShareSheet))
+            .modifier(ImageContextMenuModifier(url: url, image: image, imageSaver: imageSaver, showShareSheet: $showShareSheet))
             .sheet(isPresented: $showShareSheet) {
                 ShareSheet(activityItems: [url])
             }
+            .alert(
+                imageSaver.errorTitle,
+                isPresented: $imageSaver.error,
+                actions: {
+                    if imageSaver.errorType == .permissions {
+                        Button("Settings") {
+                            UIApplication.shared.open(URL(string: UIApplication.openSettingsURLString)!)
+                        }
+                        Button("Cancel", role: .cancel) {}
+                    }
+                },
+                message: { Text(imageSaver.errorMessage) }
+            )
     }
     
     var body: some View {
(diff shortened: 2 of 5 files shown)


jb55:
cc @bryanmontz

jb55:
On Mon, Jul 31, 2023 at 09:29:34AM -0700, doffing.brett wrote:
>> I believe this is a very non-standard way to do this. Almost every
>> app I've ever used will re-prompt for access before trying to save.
>> If we know it will fail before we try and we have a way to fix it we
>> should attempt the fix first.
>
>I would agree, and it was something I looked into. Not sure if it is
>just me, and something I will look further into, but I had to change
>the permissions from the privacy photos “additions” permission, to the
>more general photos permissions, and I would wonder if that would be
>acceptable if it is indeed the case that I have to change it.

Seems weird we would need to switch to general permissions just to
prompt for the additions permission, if this is what you're saying? 🤔

brett-doffing:
It was me. I was asking for `.readWrite` with `PHPhotoLibrary.requestAuthorization(for:)` instead of `.addOnly`, and it was causing a crash. With the fact that I can check permission status, and give an alert based on that, I would ask if we should keep the `ImageSaver` class that I implemented for unsuccessful saving? I can otherwise get rid of it?

brett-doffing:
I will keep it around and try to implement a check for a disk space error.

bryanmontz:
Once you show the photos access system prompt to the user once, it will not show again. This is also true for the GPS location prompt and the notifications prompt. So showing it again is not an option.

The console explains why the app crashes when you try to request `.readWrite` permission:
> "This app has crashed because it attempted to access privacy-sensitive data without a usage description.  The app's Info.plist must contain an NSPhotoLibraryUsageDescription key with a string value explaining to the user how the app uses this data."

Overall, this PR seems like a reasonable way to handle it: try to save the image, and report to the user if it fails. We are showing the user the Save Image button in the pop-up menu, and we could already know that it will fail (using `PHPhotoLibrary.requestAuthorization(for: .addOnly)`). But I don't think we'd change the UI because of that. Still give the user the opportunity to Save Image, and handle the failure.

damus/Util/Images/ImageSaver.swift
```diff
@@ -0,0 +1,19 @@
+// ImageSaver.swift
+
+import UIKit
+
+class ImageSaver: NSObject, ObservableObject {
```
bryanmontz, line 5: Swift classes don't need to be subclasses of `NSObject` (unless used from Objective-C). And since this class doesn't have a subclass, I'd mark it with `final`, which allows the compiler to optimize things and makes it easier to debug in the future:
```swift
final class ImageSaver: ObservableObject { ... }
```

damus/Util/Images/ImageSaver.swift
```diff
@@ -0,0 +1,19 @@
...
+    func writeToPhotoAlbum(image: UIImage) {
+        UIImageWriteToSavedPhotosAlbum(image, self, #selector(saveCompleted), nil)
+    }
+
+    @objc func saveCompleted(
+        _ image: UIImage,
+        didFinishSavingWithError error: Error?,
+        contextInfo: UnsafeRawPointer
+    ) {
+        self.error = error != nil ? true : false
```
bryanmontz, line 17: There's no reason to put `? true : false` since the expression `error != nil` already evaluates to a `Bool`.
```swift
self.error = error != nil
```

damus/Views/Images/ImageContainerView.swift
```diff
@@ -35,10 +36,18 @@ struct ImageContainerView: View {
...
+            .modifier(ImageContextMenuModifier(url: url, image: image, imageSaver: imageSaver, showShareSheet: $showShareSheet))
             .sheet(isPresented: $showShareSheet) {
                 ShareSheet(activityItems: [url])
             }
+            .alert(
+                "Failed to save photo",
+                isPresented: $imageSaver.error,
+                actions: {},
+                message: {
+                    Text("Please check the photo permissions in your Settings")
```
bryanmontz, line 48: I would clarify where the settings are: Settings.app. And since this is a complete sentence, add a period at the end.
```swift
Text("Please check the photo permissions for Damus in Settings.app.")
```

bryanmontz:
We could also improve the user experience by sending them to the Damus settings view in Settings.app with this:
```swift
UIApplication.shared.open(URL(string: UIApplication.openSettingsURLString)!)
```

damus/Util/Images/ImageSaver.swift
bryanmontz: Oh, dang. You're right. My bad.

brett-doffing:
What is unclear to me at this point is whether to implement _both_ the permissions check via `PHPhotoLibrary.requestAuthorization(for: .addOnly)` **and** the handling of failures with the `ImageSaver` class, or to implement one or the other? As it stands currently in this PR, I could get rid of `ImageSaver`, implement the permissions check alone, and keep the same UI. The reason to keep `ImageSaver` would then be to handle other failure cases, functionality that did not previously, or currently exist.

jb55:
On Tue, Aug 01, 2023 at 05:33:53AM -0700, Bryan Montz wrote:
>We could also improve the user experience by sending them to the Damus settings view in Settings.app with this:
>```swift
>UIApplication.shared.open(URL(string: UIApplication.openSettingsURLString)!)
>```

I think this is what I was remembering from other apps, not re-prompting for permissions like you mentioned which is not possible.

jb55:
what's the status of this? did we come to a conclusion ?

PR #1424, opened 2023-08-02
Fix for missing bottom half of a note

cr0bar:
Strange fix, but by increasing the height of a UiTextView past the size of any legitimate content, then re-sizes back to the correct size displaying the full content.


jb55:
thanks! merged in 39dce641316e

PR #1427, opened 2023-08-03
Second attempt at fixing SelectableText

cr0bar:
In some edge cases, the inflated UiTextView didn't render properly causing a black screen which needed the user to scroll. Dropped the inflate size and now only set where selectedTextHeight is .zero, seems more reliable.

### DIFF
--- a/damus/Components/SelectableText.swift
+++ b/damus/Components/SelectableText.swift
@@ -29,7 +29,9 @@ struct SelectableText: View {
             .padding([.leading, .trailing], -1.0)
             .onAppear {
                 self.selectedTextWidth = geo.size.width
-                self.selectedTextHeight = 100000.0
+                if(self.selectedTextHeight == .zero) {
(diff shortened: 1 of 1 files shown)


cr0bar:
Been running this patch on iOS 17b4 for the past few days and not experienced any issues with blanking out content when viewing notes. Reverted the change and instantly saw high volumes of notes with the issue.

jb55:
@cr0bar I'm still seeing this issue

jb55:
I'm seeing it https://damus.io/note12wrzl9yg25g24qm8khlhx9nngmdjg7d4hzn2dalp3h5kq2uyzaqs43qlhn

@cr0bar

cr0bar:
Unable to replicate the bug with the note you've provided. Sent a video on telegram.

jb55:
apparently this is fixed on latest iOS17 beta

Issue #1474, opened 2023-08-14
Adding Zap icon ⚡️ for Eurothon hashtag


unwired01:
Adding the ⚡️ icon to Eurothon, same as in the Zapathon hashtag. 
Given the Zapathon reference, it should be an easy task. 

![IMG_1025](https://github.com/damus-io/damus/assets/133644039/abb19684-e187-43b8-adf1-1a00d19709e9)

unwired01:
@alltheseas I would like to make those changes.
However, I don't have an iOS environment, therefore I could use your help for testing.

alltheseas:
Added this to the backlog with a good first issue tag. Anyone can submit a PR. 

Low priority, given bugs needing fixing, and sprint commitments

alltheseas:
Thank you 🙏

alltheseas:
Closing as part of a backlog triage — we're narrowing focus to the most actionable issues to reduce noise for contributors.

If this is still relevant, please reopen with updated context or reproduction steps. Thanks for reporting!

PR #1477, opened 2023-08-14
Adding zap icon to Eurothon hashtag

unwired01:
Adding the Zap ⚡️ icon to the Eurothon hashtag.
I created a new property for Eurothon.
I'm using the same svg icon file, just like the Zapathon hashtag.

I couldn't test it because I don't have a macOS environment.

Issue: #1474


jb55:
sorry we're not adding any new custom hashtags for the foreseeable future

Issue #1375, opened 2023-07-19
Apostrophe (or other punctuation) after at tag 250k Sats Bounty


alltheseas:
_what happens_
When I at tag someone in a post, and I add an apostrophe (without space), the mention (purple text is broken).

![image](https://github.com/damus-io/damus/assets/64376233/d99714a8-9e89-476c-90ab-1cd2dd90aac0)

https://damus.io/note19kndzcdtxuyf0pk4uuj8xqxrd24uq26t5zhmndufc8l9p884tqyq4rmk50

shawnyeager:
I'd argue that this an irritating bug, not an enhancement. I routinely use other clients because I'm unable to use simple punctuation in Damus. I'll commit 250k sats for a fix.

danieldaquino:
I will take a look at this today, and report back with my findings (or a fix, if I can) 👍

punterwantsawhalepass:
🤞

geeknik:
Please review my patch for bounty worthiness. :)

alltheseas:
Thanks @geeknik - Will reviews PRs.

Have you read https://github.com/damus-io/damus/blob/master/docs/CONTRIBUTING.md ?

geeknik:
I have not but I probably should have, so I will. ;) I'm just used to submitting patches on GH.

danieldaquino:
I noticed @geeknik's patch is addressing a section of code that is different from the one I was working on. Since my investigation does not seem to have overlap with his patch, I decided to continue the investigation that I had started yesterday, as it could still add some value.

# Issue reproduction

**iOS:** 16.4 (iPhone 14 Pro simulator)
**Damus:** 1.6 (13) `9fb1cc5b`
**Steps:**

1. Start composing a new note
2. In the editor, tag another user
3. Backspace to get the text cursor right at the edge of the mention link
4. Start typing something
5. Post

**Expected behaviour:** New text should be placed outside the purple link
**Actual behaviour:** New text is added as part of the purple link, and new text disappears when note is posted. 

**Video:** It might be easier to see the issue on a video, so I recorded this video to demonstrate: [https://drive.google.com/file/d/1aWVjucKZPcl319WFFfBTJCAkZEFfazk6/view?usp=sharing](https://drive.google.com/file/d/1aWVjucKZPcl319WFFfBTJCAkZEFfazk6/view?usp=sharing)

I believe this was what @shawnyeager was describing.

# Root cause and fix

[...]

PR #1487, opened 2023-08-19
Update damus.c

geeknik:
Closes: #1375


jb55:
On Fri, Aug 18, 2023 at 08:00:40PM -0700, geeknik wrote:
>Possible fix for #1375

please add tests when testing for possible fixes, it will tell is if it
actually fixes it or not.

geeknik:
I'm a bad dev haha, I'll do better. ;)

PR #1520, opened 2023-08-28
Profiles in nostrdb

jb55:
This adds profiles to nostrdb and replaces our in-memory and ProfileDatabase cache with it

- Remove in-memory Profiles caches, nostrdb is as fast as an in-memory cache
- Remove ProfileDatabase and just use nostrdb directly

## TODO

- [x] add database versions for migrations
- [x] usersearch-trie needs to be re-added until we have lmdb searching


jb55:
this is a bit crashy without transactions so I've been working hard at a new branch that adds transactions, unfortunately this lead to another large rewrite of the codebase ... sigh.

jb55:
ok transactions seem to clean up all the crashes... will test some more.

jb55:
this is done

Issue #1527, opened 2023-08-31
Fix user profile display for long names and NIP-05 domains


dmnyc:
What happens:

If a user such as npub1fpcd25q2zg09rp65fglxuhp0acws5qlphpg88un7mdcskygdvgyqfv4sld has a long name, the text wraps poorly and causes alignment issues.

What should happen:

If the "Follows you" element can be repositioned below the username to the right of the npub, it will free up more space to prevent line wrapping in most cases.

![image](https://github.com/damus-io/damus/assets/3638467/4ff87561-7443-435a-be36-258ec5422b8d)
![image](https://github.com/damus-io/damus/assets/3638467/ba8363b4-5bfc-44d4-be3f-07824e184efa)

kroucis:
The current suggestion interferes with the "√ Copied" text that appears and disappears when tapping on the two boxes to copy the pubkey. I tried a version that fades out the "Follows you" text and replaces it with the "√ Copied" prompt, but it caused some pretty ugly jumpy animations. Is there a reason why we can't move the "Follows you" badge up to the profile picture area, as shown in the attached screenshot? If it ends up interfering with the Action buttons (currently Zap and DM), we can either shrink the badge OR move the other actions into an ellipses pop-over menu. Thoughts?
<img width="562" alt="Screenshot 2023-09-02 at 00 12 32" src="https://github.com/damus-io/damus/assets/1031501/35fbed15-8cf6-4e96-9d87-903939d83729">

dmnyc:
That's a good suggestion. Anywhere other than where it is that makes logical sense would be preferable.

dmnyc:
Here's the same profile on 3 different clients. Both Snort and Primal display the NIP05 address on the line below the name to avoid spacing issues. I think Damus should do the same. 

![image](https://github.com/damus-io/damus/assets/3638467/9363fd05-762d-4f37-8970-a53e69326c00)

![image](https://github.com/damus-io/damus/assets/3638467/86f23a09-5c88-46fa-a88b-1b8681d96dab)

![image](https://github.com/damus-io/damus/assets/3638467/1a1d6c9c-6d42-4273-930f-e05c66877553)

jb55:
On Sun, Sep 03, 2023 at 09:02:48AM -0700, Maybe: Daniel ?️ wrote:
>Here's the same profile on 3 different clients. Both Snort and Primal display the NIP05 address on the line below the name to avoid spacing issues. I think Damus should do the same.

agree!

[...]

PR #1534, opened 2023-09-04
[#1527] Separate NIP-05 and username/display name onto their own lines.

kroucis:
Addresses the additional requests for #1527

The profile username header now separates each name onto its own line: display name, username, and NIP-05 domain.
Screenshot 2023-09-04 at 00 46 40

For tips/bounties:
Zap name through Alby: fragrantleaf46245@getalby.com
LNURL:
`LNURL1DP68GURN8GHJ7EM9W3SKCCNE9E3K7MF0D3H82UNVWQHKVUNPVAEXZMN5D3JKZE35XCERGDG5SVQTZ`

### DIFF
--- a/damus/Views/Profile/ProfileName.swift
+++ b/damus/Views/Profile/ProfileName.swift
@@ -80,22 +80,24 @@ struct ProfileName: View {
     }
     
     var body: some View {
-        HStack(spacing: 2) {
-            Text(verbatim: "\(prefix)\(name_choice)")
-                .font(.body)
-                .fontWeight(prefix == "@" ? .none : .bold)
+        VStack(alignment: .leading, spacing: 2) {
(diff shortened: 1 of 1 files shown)


jb55:
This is ok I guess but we should really lean into the idea of "nostr
addresses" and getting rid of the verification check which is a lie.

kroucis:
> This is ok I guess but we should really lean into the idea of "nostr addresses" and getting rid of the verification check which is a lie. Reviewed-by: William Casarin ***@***.***>
> […](#)

Perhaps that should be the subject of a longer conversation or a new issue? I simply addressed the request in the issue. I have no context for the validity of the verification check or what systems might replace it.

jb55:
yes I know I was just making a comment for future work

PR #1535, opened 2023-09-04
Finnish translation

petrikaj:
Did the Finnish translation.

(Transifex was not opened up for Finnish so here it is :)


jb55:
awesome, thanks!

Issue #919, opened 2023-04-14
Low data / cellular mode [feature request]


alltheseas:
## User Story
As a Damus pleb who uses app on the go (cellular), I want to have a setting where I can turn down Damus data needs (data saver mode), so that I don't surpass my celullar carrier's data throttling threshold, thereby avoiding a poor Damus experience.

### Acceptance Criteria
1. In settings Damus user can set cellular mode on
2. By setting cellular mode, Damus automagically detects when Damus is _not_ on wi-fi, and reduces data demands
3. This setting is turned on automatically, if it detects low data mode on iOS

alltheseas:
https://github.com/damus-io/damus/issues/645 https://github.com/damus-io/damus/pull/965 
would reduce bandwidth requirements

alltheseas:
https://stackoverflow.com/questions/2593971/iphone-simulator-simulate-a-slow-connection

https://download.developer.apple.com/Developer_Tools/Additional_Tools_for_Xcode_14.3_Release_Candidate/Additional_Tools_for_Xcode_14.3_Release_Candidate.dmg

alltheseas:
![image](https://github.com/damus-io/damus/assets/64376233/7a53a1bd-4e50-431c-8a7b-69a872aa75fb)

![image](https://github.com/damus-io/damus/assets/64376233/105afbb7-669d-4251-aae6-3bacf9fa58f6)

alltheseas:
Negentropy by @jb55 should reduce data consumption

alltheseas:
Related: disable image loading

https://github.com/damus-io/damus/issues/667

alltheseas:
Bounty posted by Tetsu, Dan Wege:

alltheseas:
@robagreda your design magic is requested, sir 👀

alltheseas:
> Bounty posted by Tetsu, Dan Wege: 100k sats
> 
> 
> 
> https://damus.io/note1uygs4fzg6dtrff326stk9thfec5meqx63f6v03qpv6axtsara7js2vwx24

225k sats

https://damus.io/note1aqg5qll3vu689j4568d22qgycysweanhhm7qcj9w9aaw9wrnyu8qwyg879

alltheseas:
![image](https://github.com/damus-io/damus/assets/64376233/8bf695a6-6223-4beb-bf53-20f9097d569b)

alltheseas:
[...]

PR #1607, opened 2023-10-14
Collapse Images

ericholguin:
This PR allows users to collapse images, essentially just showing the url link instead of the image:

nostr:note1v82v8td4qmjxneg3xgkycas4mdkgakxzhtmmen8yfdzfs6he8gfqusu5jm

https://damus.io/note1v82v8td4qmjxneg3xgkycas4mdkgakxzhtmmen8yfdzfs6he8gfqusu5jm

### DIFF
--- a/damus/Models/UserSettingsStore.swift
+++ b/damus/Models/UserSettingsStore.swift
@@ -110,6 +110,9 @@ class UserSettingsStore: ObservableObject {
     @Setting(key: "always_show_images", default_value: false)
     var always_show_images: Bool
     
+    @Setting(key: "collapse_images", default_value: false)
+    var collapse_images: Bool
+    
     @Setting(key: "hide_nsfw_tagged_content", default_value: false)
     var hide_nsfw_tagged_content: Bool
 
--- a/damus/Views/NoteContentView.swift
+++ b/damus/Views/NoteContentView.swift
@@ -26,7 +26,7 @@ struct NoteContentView: View {
     
     let damus_state: DamusState
     let event: NostrEvent
-    @State var show_images: Bool
+    @State var blur_images: Bool
     let size: EventViewKind
     let preview_height: CGFloat?
     let options: EventViewOptions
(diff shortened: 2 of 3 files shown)


jb55:
On Sat, Oct 14, 2023 at 04:44:51PM -0600, ericholguin wrote:
>This PR allows users to collapse images, essentially just showing the url link instead of the image:
>
>nostr:note1v82v8td4qmjxneg3xgkycas4mdkgakxzhtmmen8yfdzfs6he8gfqusu5jm
>You can view, comment on, or merge this pull request online at:

we should wrap this into Low-data mode #919

so instead of having "collapse images" we would just have "low data
mode" and include this PR.

I will start a low-data-mode branch and include these changes.

jb55:
>+                Toggle(NSLocalizedString("Blur images", comment: "Setting to blur images"), isOn: $settings.blur_images)
>+                    .toggleStyle(.switch)
>+
>+                Toggle(NSLocalizedString("Collapse images", comment: "Setting to only show image links"), isOn: $settings.collapse_images)

please only one change per patch. I do like that the "Always show images" was renamed, but I can't merge this patch because I'm not sure "Collapse images" is the right approach just yet (low-data-mode, etc)

ericholguin:
Okay I think I fixed this

ericholguin:
As for low data mode - I would like this option in general, specifically on iPads, images are hard to view so I'd rather just collapse them, even though I don't need Low Data in that case.

jb55:
On Thu, Oct 26, 2023 at 06:47:40PM -0700, Eric Holguin wrote:
>Okay I think I fixed this

thanks, will take another look

jb55:
On Sat, Oct 14, 2023 at 04:46:58PM -0600, ericholguin wrote:
>Okay I think I fixed this

looks like it's only 1 patch now which still mixes the two changes
together. It also looks like the blur_images field was changed without
the setting getting updated, they are the inverse of each other.

ericholguin:
Closing as I messed up the commits and this needs better UX

PR #1633, opened 2023-10-22
Add simple error handling to translate.nostr.wine

nostr-wine:
We did not initially include any error handling for translate.nostr.wine. Our API returns the following errors:

400 - {“error": "Invalid request schema"}
400 - {“error": "source and target cannot be the same language"}
400 - {"error": "query is empty after preprocessing"}
402 - {"error": "API key does not exist"}
402 - {"error": "Insufficient credits"}

With this PR, if there is an error string it will be displayed it to the user instead of showing nothing. When I get a chance, I'll translate the error messages on our end based on the target_lang of the initial request so users get error feedback in their language.


jb55:
On Sat, Oct 21, 2023 at 09:47:04PM -0400, Mazin wrote:
>---
> damus/Util/Translator.swift | 5 +++--
> 1 file changed, 3 insertions(+), 2 deletions(-)
>
>diff --git a/damus/Util/Translator.swift b/damus/Util/Translator.swift
>index 3b734f488..b718bf0a9 100644
>--- a/damus/Util/Translator.swift
>+++ b/damus/Util/Translator.swift
>@@ -107,10 +107,11 @@ public struct Translator {
>         request.httpBody = try encoder.encode(body)
>
>         struct Response: Decodable {
>-            let translatedText: String
>+            let translatedText: String?
>+            let error: String?
>         }
>         let response: Response = try await decodedData(for: request)
>-        return response.translatedText
>+        return response.error ?? response.translatedText

unfortunately this is not the right approach. We will need a new return
value for this function:

enum TranslationResult {
   case error(String)
   case success(String)
}

so that the consumer of this API knows what is an error and what is not
so that an error icon and message can be displayed appropriately.

nostr-wine:
Indeed that is a better approach but requires reworking a bit more. I'll close this for now and if I have time next week I'll work on that!

PR #1652, opened 2023-10-26
User Status fixes

ericholguin:
This PR fixes the user status view so it is more useable for smaller screen devices.

Improvements include:
- no longer showing the keyboard by default so it does not hide the clear status picker
- Prevents the keyboard from moving the view
- Adds tap to dismiss keyboard 
- segmented picker was replaced to menu style since tap to dismiss cannot work simultaneously with tap to dismiss


jb55:
Looks like you need git commit --amend and move Changelog-Changed to the
commit body so that it is not in the title.

You should try creating commits without the -m argument. It can be more
than one line.

ericholguin:
Ah okay. Fixed. Thanks

PR #1767, opened 2023-11-26
update testdata URL path method

chalvern:
URL.absoluteString returns "files:///" prefix path, which causes previews error always.
URL.path would be the right method.


jb55:
Awesome, thanks!

PR #1805, opened 2023-12-09
refactor: add customizable properties to neutral button style

ericholguin:
This PR refactors the NeutralButtonStyle to take in customizable properties for different button layouts instead of creating a struct for each situation that may arise, for example a small capsule button which I needed. Also added shape representations using NeutralButtonShape for ease of use. 

neutral buttons


jb55:
Makes sense! LGTM.

PR #2037, opened 2024-03-09
Normalize requested relay ids

ynniv:
The relay pool is removing trailing slashes when adding a relay, but not when a relay is requested. This prevents valid relays from being used:

```
get_relay("wss://relay.mutinywallet.com")  => found
get_relay("wss://relay.mutinywallet.com/") => nil
```

### DIFF
--- a/damus/Nostr/Relay.swift
+++ b/damus/Nostr/Relay.swift
@@ -120,23 +120,10 @@ class Relay: Identifiable {
     }
     
     var id: String {
-        return get_relay_id(descriptor.url)
+        return descriptor.url.id
     }
-
 }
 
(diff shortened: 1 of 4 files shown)


jb55:
Shouldn't this be in the RelayURL constructor instead?

jb55:
ideally we would use RelayURL everywhere and use that as the canonical url representation, so that lookups always work everywhere.

ynniv:
Yeah, we can fold that into RelayURL. Changing the signatures of `get_relay[s]` blows up the changeset, but here's the core of that.

jb55:
Fixed in e951370a763d

PR #2188, opened 2024-04-25
Translations

tyiu:
- Import new translations for Korean, Persian, and Polish.
- Add Finnish as a supported language
- Fix localization issues, such as missing comments, grammar, redundant code, and incorrectly marked verbatim strings


jb55:
thanks!

PR #2221, opened 2024-05-06
Reduce minimum IOS version to 15

jb55:
resurrecting @kernelkind 's work on this. wip.

### DIFF
--- a/damus.xcodeproj/project.pbxproj
+++ b/damus.xcodeproj/project.pbxproj
@@ -621,6 +621,8 @@
 		D7FF94002AC7AC5300FD969D /* RelayURL.swift in Sources */ = {isa = PBXBuildFile; fileRef = D7FF93FF2AC7AC5200FD969D /* RelayURL.swift */; };
 		E02B54182B4DFADA0077FF42 /* Bech32ObjectTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = E02B54172B4DFADA0077FF42 /* Bech32ObjectTests.swift */; };
 		E04A37C62B544F090029650D /* URIParsing.swift in Sources */ = {isa = PBXBuildFile; fileRef = E04A37C52B544F090029650D /* URIParsing.swift */; };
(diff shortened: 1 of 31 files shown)


alltheseas:
### agent review findings 
```

Findings

  - High – damus/Util/Relays/RelayBootstrap.swift:74: TmpRegion(rawValue:
    Locale.current.identifier) never matches because identifier returns
    values like en_US. The new lookup therefore drops all region-specific
    bootstrap relays for every locale. Use a true region code (e.g.
    Locale.current.regionCode) or another iOS‑15–friendly replacement so those
    regional defaults still load.
  - High – damus.xcodeproj/project.pbxproj:4055: DamusNotificationService still
    advertises an iOS 16 deployment target. On an iOS 15 device the notification
    service extension won’t load, which risks breaking whatever logic you rely
    on there (e.g. rich push handling) and may even block App Store submission
    for a 15+ app. Align the extension’s deployment target—or conditionally
    strip it for 15.
  - Medium – damus/Components/Status/UserStatusView.swift:38: The iOS 15 path
    drops the context menu entirely even though the menu items API exists before
    iOS 16. Users on 15 lose the ability to long-press and open/copy the status
    URL. You can keep the menu by calling the older contextMenu { ... } overload
    and only skipping the preview: argument on <16.

  Questions / Risks

  - For the notification service: do we need its behavior on iOS 15? If so,
    we’ll have to audit that target for newer APIs and lower its deployment
    target (or provide conditional shims).
  - On the new single-line fallbacks (e.g. zap comment/status editors), are we
    okay with the UX downgrade for 15, or should we switch to a TextEditor to
    keep multi-line entry?

  Overall the changes follow a consistent pattern—NavigationBackport for stack
  navigation and #available guards for 16-only modifiers—but the region lookup
  and extension target need fixes before shipping.

```

danieldaquino:
Closing due to extended inactivity. Please feel free to reopen if needed.

Issue #2239, opened 2024-05-09
nip10 reply support


jb55:
I realize we probably need this, and I've been working on it today. adding some tests and will PR soon

alltheseas:
relates to https://github.com/damus-io/damus/issues/2022

Issue #2233, opened 2024-05-09
Replies to quoted events get the q tag and are counted as a quote repost


jb55:
We need to strip `q` tags when replying, otherwise they are counted as a quote repost.

The main issue is that our reply code simply copies all the tags over which is not good. This is probably also causing lots of nip10 issues.

PR #2241, opened 2024-05-09
nip10: marker replies

jb55:
This should drastically increase compatibility for damus replies in other clients.

Also filter non-pubkey references when replying so we don't run into the q-tag bug.

Changelog-Added: Added nip10 marker replies
Changelog-Fixed: Fixed issue where some replies were including the q tag
Fixes: 
- https://github.com/damus-io/damus/issues/2239
- https://github.com/damus-io/damus/issues/2233


jb55:
wow https://github.com/damus-io/damus/pull/2241/commits/4ac3da76121eebcf794b7ac5e8b68bbebcdf88eb fixed so many broken threads. I integrated nostrdb so that it can crawl a thread no matter if the tags are correct or not.

PR #2265, opened 2024-05-17
added accessibility labels to maintab view

alltheseas:
added four (4) accessibility labels to maintab view

### DIFF
--- a/damus/Views/MainTabView.swift
+++ b/damus/Views/MainTabView.swift
@@ -83,6 +83,10 @@ struct TabBar: View {
                 TabButton(timeline: .search, img: "magnifyingglass.circle", selected: $selected, new_events: $new_events, isSidebarVisible: $isSidebarVisible, action: action).keyboardShortcut("3")
                 TabButton(timeline: .notifications, img: "bell", selected: $selected, new_events: $new_events, isSidebarVisible: $isSidebarVisible, action: action).keyboardShortcut("4")
             }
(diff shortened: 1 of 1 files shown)


damus/Views/MainTabView.swift
```diff
@@ -83,6 +83,10 @@ struct TabBar: View {
                 TabButton(timeline: .search, img: "magnifyingglass.circle", selected: $selected, new_events: $new_events, isSidebarVisible: $isSidebarVisible, action: action).keyboardShortcut("3")
                 TabButton(timeline: .notifications, img: "bell", selected: $selected, new_events: $new_events, isSidebarVisible: $isSidebarVisible, action: action).keyboardShortcut("4")
             }
+            .accessibilityLabel(NSLocalizedString("home", comment: "Accessibility label for home button"))
+            .accessibilityLabel(NSLocalizedString("dms", comment: "Accessibility label for dms button"))
+            .accessibilityLabel(NSLocalizedString("search", comment: "Accessibility label for search button"))
+            .accessibilityLabel(NSLocalizedString("notifications", comment: "Accessibility label for notifications button"))
```
jb55, line 89: I think these need to be on the individual buttons, this would add 4 separate accessibility lables to the entire tab bar.

Ideally you should test these to confirm they are added properly. I think there is an inspector for this:

https://developer.apple.com/documentation/accessibility/accessibility-inspector

jb55 (changes requested):

Issue #2150, opened 2024-04-16
Bug: 1.9 fix reaction count


alltheseas:
**What happens**
When I count reactions, count is off in note vs reaction view.

> there seems to be something with the reactions. The note reaction count shows 7, but when I go into the reaction menu I count 18. And it shows that I liked my own note 6 times even though I didn’t. Damus 1.9(1) TestFlight

**What I expect to happen**
I expect the same reaction count across all views. I dont expect to see I liked my own note. 

**Link to noteID, npub**

https://damus.io/nevent1qqsxx2jm8z7uruzupqndq54g7vn5kggf4yy8zhj7up6pv66y4ad3eycpz4mhxue69uhhyetvv9ujumt0wd68ytnsw43qz8rhwden5te0dehhxarj9ekh2arfdeuhwctvd3jhgtnrdakszyrhwden5te0dehhxarj9emkjmn9q9chwumn8ghj7enfd36x2u3wdehhxarj9emkjmn99ah8qatzx9e85eeexeax5ctkvashgumcx43ksvnkwe68zdrpw3shgmrex4e8verhw9nk5upsw468sae5x4ax27nwweuhzenyddu8veflvfex7ctyvdshxapaw3e82efxvakx7cnpds7kzmrv49a73t

**Screenshots/video recording**

https://v.nostr.build/7GaeD.mp4

** Versions **
Damus version: 1.9 TF

alltheseas:
Also received report from the Daniel

alltheseas:
![image](https://github.com/damus-io/damus/assets/64376233/f30beedf-fb51-487a-8d53-e469bd02d807)

![image](https://github.com/damus-io/damus/assets/64376233/53f646a4-3718-448d-902b-9b2c35c6d70f)

![image](https://github.com/damus-io/damus/assets/64376233/5dd6429c-1434-4a58-ac19-be4e04ae5b06)

alltheseas:
https://damus.io/nevent1qqsxt7yt7nt3ts578wzzv3nlgt774ny93cny07q3p6ugqwgkmhngmrspp4mhxue69uhkummn9ekx7mqpzamhxue69uhhyetvv9ujucm4wfex2mn59en8j6gpzpmhxue69uhkummnw3ezuamfdejsz9nhwden5te0wfjkccte9ec8y6tdv9kzumn9wslfunxy

alltheseas:
[...]

PR #2287, opened 2024-06-05
Fix reaction events to not tag all e and p tags in the thread

tyiu:
I've also added support for reacting to replaceable events via the `a` tag, and added a `k` tag to be explicit about what kind number of event is being reacted to.

Fixes https://github.com/damus-io/damus/issues/2150
Fixes https://github.com/nostrability/nostrability/issues/48

See https://github.com/nostr-protocol/nips/blob/master/25.md
```
The reaction event SHOULD include `a`, `e` and `p` tags pointing to the note the user is
reacting to. The `p` tag allows authors to be notified. The `e` tags enables clients 
to pull all the reactions to individual events and `a` tags enables clients to seek reactions
for all versions of a replaceable event. 

The `e` tag MUST be the `id` of the note that is being reacted to. 

The `a` tag MUST contain the coordinates (`kind:pubkey:d-tag`) of the replaceable being reacted to. 

The `p` tag MUST be the `pubkey` of the event being reacted to.

The reaction event MAY include a `k` tag with the stringified kind number
of the reacted event as its value.
```

### DIFF
--- a/damusTests/LikeTests.swift
+++ b/damusTests/LikeTests.swift
@@ -10,15 +10,7 @@ import XCTest
 
 class LikeTests: XCTestCase {
 
-    override func setUpWithError() throws {
-        // Put setup code here. This method is called before the invocation of each test method in the class.
-    }
-
-    override func tearDownWithError() throws {
-        // Put teardown code here. This method is called after the invocation of each test method in the class.
-    }
-
-    func testLikeHasNotification() throws {
+    func testReactionTextNote() throws {
         let cindy = Pubkey(hex: "9d9181f0aea6500e1f360e07b9f37e25c72169b5158ae78df53f295272b6b71c")!
         let bob = Pubkey(hex: "218837fe8c94a66ae33af277bcbda45a0319e7726220cd76171b9dd1a468af91")!
         let liked = NostrEvent(content: "awesome #[0] post",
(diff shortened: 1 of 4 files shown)


damusTests/LikeTests.swift
```diff
@@ -28,9 +20,30 @@ class LikeTests: XCTestCase {
...
+        XCTAssertTrue(like_ev.referenced_pubkeys.contains(test_keypair.pubkey))
         XCTAssertEqual(like_ev.last_refid()!, id)
+
+        let dTagValue = try XCTUnwrap(liked.tags.first { $0[0].matches_char("d") })
+        let aTag = try XCTUnwrap(like_ev.tags.first { $0[0].matches_char("a") })
+        XCTAssertTrue(aTag[1].matches_str("30023:\(test_keypair.pubkey.hex()):\(dTagValue[1])"))
+
+        let kindTag = try XCTUnwrap(like_ev.tags.first(where: { $0.count >= 2 && $0[0].matches_char("k") }))
+//        FIXME(tyiu) the assertion below fails for some reason even though in a different test, I was able to assert kind 1 just fine.
+//        XCTAssertTrue(kindTag[1].matches_str("30023"))
```
tyiu, line 46: @jb55 Any idea why this assertion fails? I'm not familiar with the NostrDB stuff.

tyiu:
Oh actually, I don't think this is the right solution. We probably still want to carry over the `p` tags that were mentions. Let me close this PR and reopen once I've figured it out.

tyiu:
Hm, NIP-25 does imply that there's only one tag for `p`. Going to reopen the PR and see what you think @jb55 . Maybe it's fine that folks don't get notified about reactions to notes that they didn't author but got mentioned in. What do you think?

alltheseas:
> Maybe it's fine that folks don't get notified about reactions to notes that they didn't author but got mentioned in

This sounds correct to me

jb55:
On Wed, Jun 05, 2024 at 07:46:04PM GMT, alltheseas wrote:
>> Maybe it's fine that folks don't get notified about reactions to
>> notes that they didn't author but got mentioned in
>
>This sounds correct to me

I like this feature though? It's cool to see people liking posts that
you're mentioned in, I believe twitter does this as well.

PR #2291, opened 2024-06-09
Remove unused Trie and UserSearchCache as they have been superseded by NostrDB

tyiu:
I've moved out `Trie` into a separate MIT-licensed library called SwiftTrie so that others can use it if they want. I'm looking into using it for when we add support for emoji search:
https://github.com/tyiu/swift-trie


jb55:
thanks!

Issue #1412, opened 2023-08-01
Do not show nsfw filter


alltheseas:
## user story
As a user browsing Damus, I do not want to see nsfw in my feed by default, so that I can manage my nostr experience. 

### acceptance criteria
1. User has an option to disable the nsfw tagged posts
2. New section in appearance for controlling content filtering settings allows for setting described in 1.
3. If user follows a certain hashtag (e.g. #grownostr), and the post also includes #nsfw, AND user has disabled nsfw posts from showing do *not* show post including nsfw
4. View nsfw should be disabled by default

https://damus.io/note18uu0sv9j2wkk4n7a8qm6qhflwavxv0nmk00v05xgu56npgw4yxhsegh4hr

alltheseas:
Previous nsfw requests: 

NIP-36 https://github.com/damus-io/damus/issues/910
https://github.com/damus-io/damus/issues/73

alltheseas:
Updated story and criteria per today's discussion @danieldaquino

danieldaquino:
Thanks @alltheseas and @jb55, I am working on the simplified changes in the appearance tab.

I will post the draft I was working on before our meeting, just in case we want to reuse some of that work whenever we want to create a specialized settings page with more fine tuned controls at a later time.

![content-moderation-setting-page-draft](https://github.com/damus-io/damus/assets/24692108/b6c6d84c-30ca-4cdc-96f5-5ae852a9738b)

[content_moderation_setting_page_draft.patch](https://github.com/damus-io/damus/files/12652160/content_moderation_setting_page_draft.patch)

alltheseas:
> Thanks @alltheseas and @jb55, I am working on the simplified changes in the appearance tab.
> 
> I will post the draft I was working on before our meeting, just in case we want to reuse some of that work whenever we want to create a specialized settings page with more fine tuned controls at a later time.
> 
> ![content-moderation-setting-page-draft](https://user-images.githubusercontent.com/24692108/268742187-b6c6d84c-30ca-4cdc-96f5-5ae852a9738b.png)
> 
> [content_moderation_setting_page_draft.patch](https://github.com/damus-io/damus/files/12652160/content_moderation_setting_page_draft.patch)

Sounds good. I added the fine tuned bit, as an incremental addition https://github.com/damus-io/damus/issues/1549 .

danieldaquino:
Implemented feature!

[...]

Issue #910, opened 2023-04-12
NIP-36 Support: Sensitive Content / Content Warning / NSFW


h3zjp:
Can we support NIP-36 (Sensitive Content / Content Warning) to hide NSFW content?

Expected behavior

![https://ul.h3z.jp/AfSwVpdu.png](https://ul.h3z.jp/AfSwVpdu.png)

Tap "表示するにはクリック" on the first image to see the text on the second image.

![https://ul.h3z.jp/eDiDUCpY.png](https://ul.h3z.jp/eDiDUCpY.png)

Current Damus

![https://ul.h3z.jp/fSd8TncT.jpeg](https://ul.h3z.jp/fSd8TncT.jpeg)

The NIP-36 specification is ignored and the text is displayed as is.

We Japanese are looking forward to this feature, so please implement it.
Best regards.

alltheseas:
Can you provide a noteID to the above example?

alltheseas:
https://github.com/nostr-protocol/nips/blob/master/36.md

h3zjp:
I don't understand what noteID refers to, but I will include the JSON query that was POST to the relay.

![https://ul.h3z.jp/zlHDsCl8.png](https://ul.h3z.jp/zlHDsCl8.png)
![https://ul.h3z.jp/gs6RMW1s.png](https://ul.h3z.jp/gs6RMW1s.png)
![https://ul.h3z.jp/UjzwhYDT.png](https://ul.h3z.jp/UjzwhYDT.png)
![https://ul.h3z.jp/KzlJOYYa.png](https://ul.h3z.jp/KzlJOYYa.png)

h3zjp:
The client used in the test is called "Rabbit".
https://github.com/syusui-s/rabbit

The implementation of this client displays the message "表示するにはクリック" and below it the text written in the NIP-36 content-warning tag.

alltheseas:
> I don't understand what noteID refers to,

Can you send a link to the note?

Here is an example link to a random note:

https://damus.io/note1ha6v4zp77tvnnptv6muxnjj6al8sqm7uumcwff3qz0a6a0l4tf9qp9gucf

h3zjp:
I see! So that's what you meant!
I'm sorry.

This is it!
[...]

PR #2294, opened 2024-06-12
Hide NIP-36 notes as #nsfw

kunigaku:
Filters posts with the NIP-36 "content-warning" tag in the same way as the #nsfw hashtag.
https://github.com/nostr-protocol/nips/blob/master/36.md

Related issues: #910, #1412

Screenshots:
Show #nsfw and NIP-36
Show #nsfw and NIP-36
Hide #nsfw and NIP-36
Hide #nsfw and NIP-36

This is a test post.
https://damus.io/note1z0jcu9wgevaajcvk947ea2aj8wv9nxlttj5dvj6wxfyz0v5ehpnq8t9l2p

The post on Rabbit.
https://rabbit.syusui.net/#/npub1l6gzvpjar07swqqq9gg9eshgtvtp2tyynv405jrsgxwrhpcupgcqqvy0hm
Rabbit Screenshot hide NIP-36
Rabbit Screenshot show NIP-36

### DIFF
--- a/damus/Models/ContentFilters.swift
+++ b/damus/Models/ContentFilters.swift
@@ -25,7 +25,7 @@ enum FilterState : Int {
 
 /// Simple filter to determine whether to show posts with #nsfw tags
 func nsfw_tag_filter(ev: NostrEvent) -> Bool {
-    return ev.referenced_hashtags.first(where: { t in t.hashtag == "nsfw" }) == nil
+    return ev.referenced_hashtags.first(where: { t in t.hashtag == "nsfw" }) == nil && References<ContentWarningTag>(tags: ev.tags).first == nil
 }
 
 func get_repost_of_muted_user_filter(damus_state: DamusState) -> ((_ ev: NostrEvent) -> Bool) {
--- a/damus/Nostr/Id.swift
+++ b/damus/Nostr/Id.swift
@@ -130,6 +130,29 @@ struct ReplaceableParam: TagConvertible {
     var keychar: AsciiCharacter { "d" }
 }
 
+struct ContentWarningTag: TagConvertible {
+    let reason: String?
+
+    static func from_tag(tag: TagSequence) -> ContentWarningTag? {
+        var i = tag.makeIterator()
+
+        guard tag.count >= 1,
+              let t0 = i.next(),
+              t0.matches_str("content-warning") else {
+            return nil
+        }
+        
+        guard let t1 = i.next() else {
+            return ContentWarningTag(reason: nil)
+        }
+
+        return ContentWarningTag(reason: t1.string())
+    }
+    var tag: [String] {
+        self.reason == nil ? ["content-warning"] : ["content-warning", self.reason!]
+    }
+}
+
 struct Signature: Hashable, Equatable {
     let data: Data
 


damus/Models/ContentFilters.swift
```diff
@@ -25,7 +25,7 @@ enum FilterState : Int {
 
 /// Simple filter to determine whether to show posts with #nsfw tags
 func nsfw_tag_filter(ev: NostrEvent) -> Bool {
-    return ev.referenced_hashtags.first(where: { t in t.hashtag == "nsfw" }) == nil
+    return ev.referenced_hashtags.first(where: { t in t.hashtag == "nsfw" }) == nil && References<ContentWarningTag>(tags: ev.tags).first == nil
```
jb55, line 28: dont we need to check the reason == nsfw in the content warning case?

kunigaku: We should consider any post with the "content-warning" tag as sensitive.
Therefore, there is no need to check the "reason."

kunigaku: This is a post without reason.
https://nostter.app/nevent1qqswynxvz0wth52wjq3xyq7l795q3w0fe83ym0ewetnx8u9gslrhmhswwl5pw
```JSON
{
  "content": "に",
  "created_at": 1718237816,
  "id": "e24ccc13dcbbd14e90226203dff16808b9e9c9e24dbf2ecae663f0a887c77dde",
  "kind": 1,
  "pubkey": "1a35b54ef7752af54cacbeedf0f349e320f0a2ee50142883134c3ee31879ce71",
  "sig": "3a1dc4873e2a8527ecfce2526af8701702e0d1bb3b4c4ecf45e8026cae27ceb84f40144e8409ca4a325e8e93f39fcba75ddf1c7ad0e57ba53f7e538a4da92b2a",
  "tags": [
    [
      "content-warning"
    ]
  ]
}
```

jb55:
On Wed, Jun 12, 2024 at 05:22:59PM GMT, kunigaku wrote:
>@kunigaku commented on this pull request.
>
>
>
>> @@ -25,7 +25,7 @@ enum FilterState : Int {
>
> /// Simple filter to determine whether to show posts with #nsfw tags
> func nsfw_tag_filter(ev: NostrEvent) -> Bool {
>-    return ev.referenced_hashtags.first(where: { t in t.hashtag == "nsfw" }) == nil
>+    return ev.referenced_hashtags.first(where: { t in t.hashtag == "nsfw" }) == nil && References<ContentWarningTag>(tags: ev.tags).first == nil
>
>We should consider any post with the "content-warning" tag as sensitive.
>Therefore, there is no need to check the "reason."

hmm... the way content-warning was used on the fediverse is that you could
mark a message as "politics", so if you were not interested in politics
you could skip that.

Our filter is for NSFW content, if we were filtering political content
with this it would be incorrect.

mattn:
Non-developers are unlikely to comment on GitHub. But I know many Nostr users who would love to have NIP-36. And they are unfortunately resigned to the fact that Damus does not have it.

mattn:
IMO, whatever the reason, I think the contributor's consideration that "maybe some people don't want to see it" should be taken into account.

TsukemonoGit:
When posting images that may be objectionable to a certain number of people, a content-warning may be added as a warning (e.g., because of insects), but a content-warning added by another client's user out of consideration for others is not effective for Damus users, causing them to feel uncomfortable.

jb55:
On Sat, Jun 15, 2024 at 12:15:42AM GMT, mono wrote:
>When posting images that may be objectionable to a certain number of
>people, a content-warning may be added as a warning (e.g., because of
>insects), but a content-warning added by another client's user out of
>consideration for others is not effective for Damus users, causing them
>to feel uncomfortable.

I'm not against adding support for content-warning, I just think we need
a proper control that hides content with a collpasible control on the
note without hiding the entire note. This nsfw filter is not the right
for this.

The nsfw filter will hide the entire note from the timeline. If we did
this for content warnings people would say we are shadow-banning people.

jb55:
On Sat, Jun 15, 2024 at 09:46:13AM GMT, William Casarin wrote:
>On Sat, Jun 15, 2024 at 12:15:42AM GMT, mono wrote:
>>When posting images that may be objectionable to a certain number of
>>people, a content-warning may be added as a warning (e.g., because of
>>insects), but a content-warning added by another client's user out of
>>consideration for others is not effective for Damus users, causing them
>>to feel uncomfortable.
>
>I'm not against adding support for content-warning, I just think we need
>a proper control that hides content with a collpasible control on the
>note without hiding the entire note. This nsfw filter is not the right
>for this.
>
>The nsfw filter will hide the entire note from the timeline. If we did
>this for content warnings people would say we are shadow-banning people.

Alternatively we should have another filter setting for filtering all
content warnings instead of re-using the nsfw filter.

TsukemonoGit:
I think a proper control that hides content with a collapsible control on the note without hiding the entire note would be a good idea.

danieldaquino:
Due to a high volume of PRs, I am closing older inactive PRs to make our list more manageable. Please feel free to reopen as needed.

Thank you for your understanding.

PR #2303, opened 2024-06-21
Add Apple translation popovers for notes for iOS 17.4+ and macOS 14.4+

tyiu:
Changelog-Added: Add Apple translation popovers for notes for iOS 17.4+ and macOS 14.4+

https://github.com/damus-io/damus/assets/963907/4fdddb61-9cd0-4514-80fe-79298d9c92a6

Simulator Screenshot - iPhone 15 Pro - 2024-07-16 at 13 07 06 Medium

These changes are gated around iOS and macOS versions. If an older OS is being run, Apple translations won't be offered.

### DIFF
--- a/damus/Models/TranslationService.swift
+++ b/damus/Models/TranslationService.swift
@@ -38,7 +38,13 @@ enum TranslationService: String, CaseIterable, Identifiable, StringCodable {
     var model: Model {
         switch self {
         case .none:
-            return .init(tag: self.rawValue, displayName: NSLocalizedString("none_translation_service", value: "None", comment: "Dropdown option for selecting no translation service."))
+            let displayName: String
+            if TranslationService.isAppleTranslationPopoverSupported {
+                displayName = NSLocalizedString("apple_translation_service", value: "Apple", comment: "Dropdown option for selecting Apple as a translation service.")
+            } else {
+                displayName = NSLocalizedString("none_translation_service", value: "None", comment: "Dropdown option for selecting no translation service.")
+            }
+            return .init(tag: self.rawValue, displayName: displayName)
         case .purple:
             return .init(tag: self.rawValue, displayName: NSLocalizedString("Damus Purple", comment: "Dropdown option for selecting Damus Purple as a translation service."))
         case .libretranslate:
@@ -49,6 +55,14 @@ enum TranslationService: String, CaseIterable, Identifiable, StringCodable {
             return .init(tag: self.rawValue, displayName: NSLocalizedString("NoKYCTranslate.com (Prepay with BTC)", comment: "Dropdown option for selecting NoKYCTranslate.com as the translation service."))
          case .winetranslate:
             return .init(tag: self.rawValue, displayName: NSLocalizedString("translate.nostr.wine (DeepL, Pay with BTC)", comment: "Dropdown option for selecting translate.nostr.wine as the translation service."))
+        }
+    }
+
+    static var isAppleTranslationPopoverSupported: Bool {
+        if #available(iOS 17.4, macOS 14.4, *) {
+            return true
+        } else {
+            return false
         }
     }
 }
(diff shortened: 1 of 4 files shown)


alltheseas:
@tyiu is it possible to set up auto translate using iOS offline translations? That is, remove the need to tap "translate  note"

tyiu:
> @tyiu is it possible to set up auto translate using iOS offline translations? That is, remove the need to tap "translate  note"

Technically yes, but only in iOS 18.0+

jb55:
very cool, will test this soon

tyiu:
There's a bug when `damus_state.settings.auto_translate` is set to true but `Offline Translations` is selected. It won't show the Translate button in that case, which is incorrect. I'll work on a fix, and also look into the auto translations with iOS 18.0.

tyiu:
I mischaracterized the this PR. It's not exclusively offline note translations from Apple. The translation requests can still go to Apple servers, but can be offline as well via downloaded languages if there's network connectivity issues. I've updated the description, commit, and PR title accordingly

>There's a bug when damus_state.settings.auto_translate is set to true but Offline Translations is selected. It won't show the Translate button in that case, which is incorrect. I'll work on a fix, and also look into the auto translations with iOS 18.0
I fixed the bug that I reported in my previous comment.

damus/Models/TranslationService.swift
```diff
@@ -38,7 +38,13 @@ enum TranslationService: String, CaseIterable, Identifiable, StringCodable {
     var model: Model {
         switch self {
         case .none:
-            return .init(tag: self.rawValue, displayName: NSLocalizedString("none_translation_service", value: "None", comment: "Dropdown option for selecting no translation service."))
+            let displayName: String
+            if TranslationService.isAppleTranslationPopoverSupported {
```
tyiu, line 42: I'd like to solicit feedback here. Right now, if you have iOS 17.4+, there's no way to turn off translations. It's just using Apple by default, but requires an explicit tap on the Translate Note button.

Alternatively, I could make `None` and `Apple` separate options, but it feels like we have an opportunity to be more opinionated here to just say everyone needs translations.

danieldaquino: I think we need to consider the privacy aspect, given that some of those translate requests might be sent to Apple servers.

In my opinion, we should be opinionated and place the button "Translate note" by default, and then add a confirmation dialog explaining the user the text _might_ be sent to Apple servers, with an option to dismiss that dialog forever if the user chooses to do so.

For the iOS 18 feature where translations can be made automatically, I believe that should be opt-in.

tyiu: > I think we need to consider the privacy aspect, given that some of those translate requests might be sent to Apple servers.
> 
> In my opinion, we should be opinionated and place the button "Translate note" by default, and then add a confirmation dialog explaining the user the text _might_ be sent to Apple servers, with an option to dismiss that dialog forever if the user chooses to do so.

Simulator Screenshot - iPhone 16 Pro - 2024-09-22 at 00 44 41

It turns out that Apple shows exactly that confirmation dialog on the first attempt, so I think we're OK there. Agreed about the automatic iOS 18 translations feature, though.

tyiu:
I rebased this branch on the latest master and force pushed. No other changes were made.

damus/Models/TranslationService.swift
danieldaquino: Thank you @tyiu, I will take a second look at this PR soon with this in mind!

danieldaquino (approved):
Thank you @tyiu, reviewed and tested it!

- Default settings and the overall flow makes a lot of sense
- Works well _(Tested on iPhone 13 mini on iOS 18.0 switching between Damus Purple translation and Apple)_
- I really like how clear and readable the code changes are! Good job!

Merging!

Issue #2318, opened 2024-07-02
Design: add Tor relay icon in relay view


alltheseas:
It is difficult to distinguish Tor vs clearnet relays.

alltheseas:
https://imgs.search.brave.com/AIGM4kCKkI0GPwHvMY3A0LNzE9KHXnpXst6SCIqDbVQ/rs:fit:500:0:0:0/g:ce/aHR0cHM6Ly91cGxv/YWQud2lraW1lZGlh/Lm9yZy93aWtpcGVk/aWEvY29tbW9ucy9j/L2M5L1Rvcl9Ccm93/c2VyX2ljb24uc3Zn

https://upload.wikimedia.org/wikipedia/commons/c/c9/Tor_Browser_icon.svg

alltheseas:
@ericholguin

alltheseas:
![image](https://github.com/damus-io/damus/assets/64376233/5ea16dd5-15c8-48a7-8e43-11d893ae5a0f)

PR #2319, opened 2024-07-03
relay: Add Tor Relay Image

ericholguin:
This PR just adds the tor icon to relays ending with .onion

Changelog-Added: Tor relay icon

Closes: #2318


alltheseas:
@danieldaquino very straightforward PR for your consideration for 1.9

damus/Views/Relays/RelayPicView.swift
```diff
@@ -42,19 +42,24 @@ struct InnerRelayPicView: View {
...
-                        view.framePreloadCount = 3
-                    }
-                    .placeholder { _ in
-                        Placeholder(url: url)
-                    }
-                    .scaledToFit()
+                if url.absoluteString.contains(".onion") {
+                   Image("tor")
+                        .scaledToFit()
+                } else {
```
danieldaquino, line 48: Should we add the Tor icon in addition to the relay image (like a badge, for example), instead of replacing the relay icon?

damus/Views/Relays/RelayPicView.swift
```diff
@@ -42,19 +42,24 @@ struct InnerRelayPicView: View {
...
-                    }
-                    .cancelOnDisappear(true)
-                    .configure { view in
-                        view.framePreloadCount = 3
-                    }
-                    .placeholder { _ in
-                        Placeholder(url: url)
-                    }
-                    .scaledToFit()
+                if url.absoluteString.contains(".onion") {
```
danieldaquino, line 45: Although unlikely, this logic might match other non-onion URLs (for example `wss://relay.onionlovers.com/v2`). We can improve this logic to avoid such mismatches, by using properties such as [the host property in URLComponents](https://developer.apple.com/documentation/foundation/urlcomponents/1779882-host) and [hasSuffix](https://developer.apple.com/documentation/swift/string/hassuffix(_:)), for example.

danieldaquino (commented):
Looks mostly good, added a few comments

damus/Views/Relays/RelayPicView.swift
alltheseas: Yes, this makes more sense as relays have a unique icon. This information serves to more quickly communicate which relays a user is subscribed to.

danieldaquino:
Thank you @ericholguin, changes look good! We are just missing the sign-off line in the commit message. Can we please add that? _(or if you authorize me, I can add that on your behalf)_

PR #2334, opened 2024-07-15
flatbuffers: don't crash if there are flatbuffer errors

jb55:
Some japanese user profiles are breaking the flatbuffer profile builder for some reason.

This is a pretty bad crash, we'll need to push this asap

Changelog-Fixed: Fix pretty bad crash when building flatbuffer profiles


jb55:
applied in 616f730ae5cc

PR #2335, opened 2024-07-15
Fix a few crashes on 1.9

jb55:
Some japanese user profiles are breaking the flatbuffer profile builder for some reason

Changelog-Fixed: Fix pretty bad crash when building flatbuffer profiles

### DIFF
--- a/damus/Models/MutelistManager.swift
+++ b/damus/Models/MutelistManager.swift
@@ -111,12 +111,16 @@ class MutelistManager {
     private func add_mute_item(_ item: MuteItem) {
         switch item {
         case .user(_, _):
+            guard !users.contains(item) else { return }
             users.insert(item)
         case .hashtag(_, _):
+            guard !hashtags.contains(item) else { return }
             hashtags.insert(item)
         case .word(_, _):
+            guard !words.contains(item) else { return }
             words.insert(item)
         case .thread(_, _):
+            guard !threads.contains(item) else { return }
             threads.insert(item)
         }
     }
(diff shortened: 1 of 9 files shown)


jb55:
Went through and fixed a bunch more crashes

nostrdb/flatcc/flatcc_builder.h
```diff
@@ -82,7 +82,7 @@ extern "C" {
  * Note: some internal assertion will remain if disabled.
  */
 #ifndef FLATCC_BUILDER_ASSERT_ON_ERROR
-#define FLATCC_BUILDER_ASSERT_ON_ERROR 1
+#define FLATCC_BUILDER_ASSERT_ON_ERROR 0
```
danieldaquino, line 85: Have we ruled out the risk that the aforementioned errors could corrupt flatbuffer data and cause further issues down the line?

damus/Models/MutelistManager.swift
danieldaquino, line 1: The changes in this file LGTM

nostrdb/flatcc/flatcc_builder.h
jb55: it is handled here:

```
	NdbProfile_parse_json_table(ctx, buf, buf + bufsiz, profile);
	if (ctx->error)
		return 0;
```

in `ndbprofile_parse_json`

jb55:
Closing in favor of #2336

nostrdb/flatcc/flatcc_builder.h
danieldaquino: Nice!

Do we have similar guards around other flatbuffer objects like ndbnotes and such (given that this constant looks global)? If we do, then this LGTM

PR #2376, opened 2024-08-09
ux: Profile Edit Improvements

ericholguin:
This PR adds improvements to the profile edit view.
The banner image is changed from the old ostrich image to the fresh new damoose.
The image and banner url text entries have been removed from the edit form and now live under the image selector menu.
Selecting the Image URL menu option presents a sheet where a user can update the image URL.
There are now safe guards in place for users who update their profile, if they make any changes and try to navigate back to home they will
get an alert asking if they want to discard changes. The Save button is also more prominent.

Changelog-Changed: Changed the default banner from ostriches to damoose
Changelog-Added: Added profile edit safe guards
Changelog-Changed: Changed image and banner url text fields to new sheet view

Signed-off-by: ericholguin <ericholguin@apache.org>


danieldaquino (approved):
Thank you @ericholguin, changes look good! The profile edit UX looks much better now!

I will squash the commits into one when merging and use the message of the second commit

Issue #2323, opened 2024-07-04
Reposts occasionally show unformatted


jb55:

jb55:
Looks like this is still an issue

danieldaquino:
I will start from the premise that most lack of formatting on push notifications have the same root cause across all types of notifications (not just repost notifications)

danieldaquino:
I am seeing two patterns of Ndb initialization error, usually when lmdb loads and memory maps the db file
- When Damus is open, it usually fails with error code "1" which is "Operation not permitted". It's probably the same reason as we have seen in the highlighter extension
- When Damus is closed, it usually fails with error code "12" which is "Cannot allocate memory". It's probably trying to allocate more memory than is allowed by the extension. Some sources say our memory limit for the extension is as little as 25 MB (https://stackoverflow.com/questions/50764003/what-memory-limit-for-apple-notifications-service-extension)

I will experiment with setting a lower memory map size

danieldaquino:
> * When Damus is closed, it usually fails with error code "12" which is "Cannot allocate memory". It's probably trying to allocate more memory than is allowed by the extension. Some sources say our memory limit for the extension is as little as 25 MB (https://stackoverflow.com/questions/50764003/what-memory-limit-for-apple-notifications-service-extension)

Now I am seeing some instances where it succeeds, and the `mdb_env_open failed, error 12` error log appears 3 times before succeeding, which indicates it could allocate about 4GB at that time 🤔

danieldaquino:
I can't repro the memory error any longer, I will investigate the other one

jb55:
we can try reducing the number of ingest threads to 1 or 0 to see if that fixes memory alloc issues, these are the only things that really allocate memory (for ingest queues).

jb55:
oh wait 12 is an lmdb message, so maybe extensions can't do large virtual memory mapping? hmm.

on damus we have a special large virtual memory entitlement, should make sure that we have that on the extension as well.

jb55:
I just read this in the LMDB docs:

```
int  mdb_env_create(MDB_env **env);

	/** @brief Open an environment handle.
	 *
	 * If this function fails, #mdb_env_close() must be called to discard the #MDB_env handle.
```

not sure if this is related, but we don't actually do that! So maybe subsequent smaller mapsize calls are failing when they don't need to (we keep reducing mapsize in a loop until it succeeds).

[...]

PR #2410, opened 2024-09-01
Fix various issues with nostrdb and our notifications extension

jb55:
I've added the extended virtual memory entitlement so we can even open nostrdb in our notification extension.

We also have a commit that attempts to fix various issues when acquiring a IPC semaphore on iOS.

See: https://github.com/damus-io/damus/issues/2323#issuecomment-2323181204

Running this patch gives us these names:

mdb_env_setup_locks: using semnames
  'group.com.damus/MDBrwDDi_FHxD' (29),
  'group.com.damus/MDBwwDDi_FHxD' (29)

From old Apple docs:

> IPC and POSIX Semaphores and Shared Memory
>
> Normally, sandboxed apps cannot use Mach IPC, POSIX semaphores and
> shared memory, or UNIX domain sockets (usefully). However, by specifying
> an entitlement that requests membership in an application group, an app
> can use these technologies to communicate with other members of that
> application group.
>
> Note: System V semaphores are not supported in sandboxed apps.
>
> UNIX domain sockets are straightforward; they work just like any other
> file.
>
> Any semaphore or Mach port that you wish to access within a sandboxed
> app must be named according to a special convention:
>
> POSIX semaphores and shared memory names must begin with the application
> group identifier, followed by a slash (/), followed by a name of your
> choosing.
>
> Mach port names must begin with the application group identifier,
> followed by a period (.), followed by a name of your choosing.
>
> For example, if your application group’s name is
> Z123456789.com.example.app-group, you might create two semaphores named
> Z123456789.myappgroup/rdyllwflg and Z123456789.myappgroup/bluwhtflg. You
> might create a Mach port named
> Z123456789.com.example.app-group.Port_of_Kobe.
>
> Note: The maximum length of a POSIX semaphore name is only 31 bytes, so
> if you need to use POSIX semaphores, you should keep your app group
> names short.

Link: https://github.com/damus-io/damus/issues/2323#issuecomment-2323305949


jb55:
This is in master and release-1.10 now

PR #2411, opened 2024-09-01
Highlighter entitlements

jb55:
This is 

- #2383

With extended virtual addressing entitlements and lmdb fixes to attempt to fix lmdb issues


jb55:
highlighter seems to be working ok for me now

PR #2426, opened 2024-09-06
A few notification fixes

jb55:
there are many states in which it will just outright fail to format unnecessarily. For example, if there is no display_name (which might not be there, so we shouldn't fail outright on these). This PR fixes that.

We also accept kind1 notes where you are tagged but where there are no mentions in the contents. We call these "Tagged" notifications.

### DIFF
--- a/DamusNotificationService/NotificationService.swift
+++ b/DamusNotificationService/NotificationService.swift
@@ -27,9 +27,9 @@ class NotificationService: UNNotificationServiceExtension {
         // Log that we got a push notification
         Log.debug("Got nostr event push notification from pubkey %s", for: .push_notifications, nostr_event.pubkey.hex())
         
-        guard let state = NotificationExtensionState(),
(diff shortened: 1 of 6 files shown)


DamusNotificationService/NotificationService.swift
```diff
@@ -39,7 +39,11 @@ class NotificationService: UNNotificationServiceExtension {
             contentHandler(improved_content)
             return
         }
-        
+
+        let txn = state.ndb.lookup_profile(nostr_event.pubkey)
+        let profile = txn?.unsafeUnownedValue?.profile
+        let name = Profile.displayName(profile: profile, pubkey: nostr_event.pubkey).displayName
```
danieldaquino, line 45: Nice catch!

danieldaquino:
@jb55 thank you, changes look good to me! Before merging can I change the target branch to `release_1.10` and then I merge `release_1.10` back to `master`?

danieldaquino:
Rebased and pushed to `release_1.10` manually, and merged up to `master`

Commits `d9bbca1005721788e98f1dc954a8f0c5b5c44943` up to `cb241741e350222160f7095487a7310ec3051bfb`

Issue #2373, opened 2024-08-08
iOS 18 - Opening Video player changes theme to black, and can't be switched afterwards


jb55:

alltheseas:
Jack reported dark mode during day randomly. Unclear if Jack is using the video player at this time. 

I suspect Jack is on iOS 18 beta.

https://damus.io/note1teu2tgvluv4p3q2hme2axd73jqcsu75xnl6xncfuz3rrfga72f3q6eggec

alltheseas:
@cr0bar to investigate

jb55:
this is so random. like seriously wtf ios

jb55:
> I suspect Jack is on iOS 18 beta.

he is according to internal testflight logs

danieldaquino:
Fixed and merged to 1.10: https://github.com/damus-io/damus/commit/6fa2e8b5c635a31e765b908e94b370b0b9b88fb3

danieldaquino:
cc @jb55

Issue #2342, opened 2024-07-17
Image Carousel broken iOS 18 beta


alltheseas:
Reported by @tyiu 

> Is the image carousel broken for anyone? I’m on iOS 18 beta 3. Tapping on an image doesn’t bring up the carousel.

> Seems like this was happening only on the latest build from master, not TestFlight.

danieldaquino:
This is still an issue on the official (non-beta) iOS 18.0 release

**Device:** iPhone 13 mini (physical device)
**iOS:** 18.0
**Damus:** Current tip of 1.10 `dfa72fceb10dc0cd6867870acfc1f2b69c890722`

danieldaquino:
Implemented https://github.com/damus-io/damus/pull/2498 as a tentative PR which improves the iOS 18 situation. Not a final fix, but improves the behavior here somewhat. 

I will keep investigating to find a better fix.

PR #2473, opened 2024-09-13
Fix for iOS 18 - Opening Video player changes theme to black, and can…

cr0bar:
…'t be switched afterwards #2373

Removed line which forces preferred colour scheme to dark.

### DIFF
--- a/damus/Views/NoteContentView.swift
+++ b/damus/Views/NoteContentView.swift
@@ -121,7 +121,6 @@ struct NoteContentView: View {
                 .padding(.top)
         }
         .background(.thinMaterial)
-        .preferredColorScheme(.dark)
         .onTapGesture(perform: {
             damus_state.nav.push(route: Route.Thread(thread: .init(event: self.event, damus_state: damus_state)))
             dismiss()
--- a/damus/Views/Purple/DamusPurpleAccountView.swift
+++ b/damus/Views/Purple/DamusPurpleAccountView.swift
@@ -64,7 +64,6 @@ struct DamusPurpleAccountView: View {
                 .padding(.bottom, 20)
             }
             .foregroundColor(.white.opacity(0.8))
-            .preferredColorScheme(.dark)
             .background(.ultraThinMaterial, in: RoundedRectangle(cornerRadius: 12, style: .continuous))
             .padding()
         }


cr0bar:
Same bug also applies to DamusPurpleAccountView.swift, only spotted it after the first commit.

jb55:
that was simple! thanks!

danieldaquino:
@cr0bar, have you tested if those views still look as expected after making these changes? 

I remember I had to apply dark mode on the purple view for some reason

cr0bar:
Everything looked ok when testing in the simulator. I'll push a build to my physical and run it for a few days to check. If it's an issue I'll have to add a dirty fix to temp the current colour scheme on load and set it back when exiting.

danieldaquino:
Thank you @cr0bar. I tried running it on my device and I found a few aesthetic differences on the purple page when using light mode. They are a bit subtle, but I am pasting some screenshots to present what I have seen

| Expected look | Current look (in light mode) |
|--------|--------|
| IMG_2625 | IMG_2626 |

The list background color is probably acceptable if we can't easily fix it, but the date next to the star needs a foreground color override in this view to avoid legibility issues.

danieldaquino:
For some reason I can't open the full screen image carousel on iOS 18 to test light mode with these changes. I think there might be another issue (unrelated to this PR) going on. I will look into it.

danieldaquino:
> For some reason I can't open the full screen image carousel on iOS 18 to test light mode with these changes. I think there might be another issue (unrelated to this PR) going on. I will look into it.

Ok, that one is known iOS 18 issue (https://github.com/damus-io/damus/issues/2342). I will test with iOS 17 for now

danieldaquino:
I checked the full screen carousel as well. I don't mind the new light mode look, but it seems to have legibility issues as well

| How it looks like before the changes | How it looks after changes (in light mode) |
|--------|--------|
| Simulator Screenshot - iPhone SE (3rd generation) - 2024-09-18 at 12 15 50 | Simulator Screenshot - iPhone SE (3rd generation) - 2024-09-18 at 12 13 09 |

**Tested on:** iPhone simulator, iOS 17.5


@cr0bar, can you please take a look at those?

danieldaquino:
@cr0bar, I made the remaining fixes on top of your commit so that we can include your fix on the 1.10 release that is scheduled this week.

The modified co-authored commit was merged into `release_1.10` (commit: https://github.com/damus-io/damus/commit/6fa2e8b5c635a31e765b908e94b370b0b9b88fb3), and that will be merged back to the `master` branch soon.

Thanks!

PR #2497, opened 2024-09-19
Don't reconnect when we don't have to

jb55:
We are reconnecting multiple times for two separate reasons:

1. On a cancellation "error" which does not warrant a reconnect

2. In our reconnection backoff it doesn't check If we are already
connecting or connected. We check this so we don't reconnect multiple
times.

This fixes many reconnection issues and makes Damus feel wayyy snappier.

### DIFF
--- a/damus/Nostr/RelayConnection.swift
+++ b/damus/Nostr/RelayConnection.swift
@@ -46,9 +46,10 @@ final class RelayConnection: ObservableObject {
             
             if err == nil {
                 self.last_pong = .now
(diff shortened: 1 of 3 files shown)


danieldaquino:
@jb55, would you like to include this to 1.10? If so we should rebase the PR to target `release_1.10`

jb55:
Oh right, will do

jb55:
Merged in 51ee4046a0471369485c4f9838e1958d56245b7d

PR #2499, opened 2024-09-22
Don't reconnect when we don't have to (1.10)

jb55:
This is

- #2497 

but applied to 1.10


jb55:
merged in 1e85bb946d14 (I had this PR targeting the wrong base)

PR #2580, opened 2024-10-11
multiple images upload

scoder1747:
## Summary

This PR allows selecting multiple images option in the Media Picker Sheet. The selected images are uploaded using the default `handle_upload()` method while performed in an array iteration. 

The motivation comes from the fact that it is often tedious to select one image at a time from the sheet and repeat the same process for multiple uploads.

The demo of the feature is attached here:

https://github.com/user-attachments/assets/f3c4eb7e-02a0-4f7a-9741-60b34cdbbce3

## Checklist

- [ ] I have read (or I am familiar with) the [Contribution Guidelines](../docs/CONTRIBUTING.md)
- [ ] My PR is either small, or I have split it into smaller logical commits that are easier to review
- [ ] I have added the signoff line to all my commits. See [Signing off your work](../docs/CONTRIBUTING.md#sign-your-work---the-developers-certificate-of-origin)
- [ ] I have added appropriate `Closes:` or `Fixes:` tags in the commit messages wherever applicable, or made sure those are not needed. See [Submitting patches](https://github.com/damus-io/damus/blob/master/docs/CONTRIBUTING.md#submitting-patches)

### DIFF
--- a/damus/Views/MediaPicker.swift
+++ b/damus/Views/MediaPicker.swift
@@ -117,7 +117,7 @@ struct MediaPicker: UIViewControllerRepresentable {
         var configuration = PHPickerConfiguration(photoLibrary: .shared())
         configuration.selectionLimit = 0 // Allows multiple media selection
         configuration.filter = imagesOnly ? .images : .any(of: [.images, .videos])
-        
+        configuration.selection = .ordered // images are returned in the order they were selected + numbered badge displayed
         let picker = PHPickerViewController(configuration: configuration)
         picker.delegate = context.coordinator as any PHPickerViewControllerDelegate
         return picker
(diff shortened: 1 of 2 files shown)


jb55:
I tried this, but when I selected 2 images the progress bar completed, and it took another 30 seconds for the second image to appear. This was pretty confusing.

scoder1747:
@jb55 I have addressed the issue you faced in subsequent commits

jb55:
it works, and I will merge it. I would like to see a followup that ensures the order in which they were clicked. right now its all async and the images come in with random order

scoder1747:
@jb55 I have resolved the issue about the order in the new commit

scoder1747:
Screenshot 2024-10-16 at 9 52 41 AM

jb55:
@scoder1747 this adds a ordered indicator but it does not ensure that the order is correct when adding the elements to the note. the order would have to be saved somewhere and then the URLs and URL previews would have to be inserted in the correct locations. We can scope this to another PR though.

I have a feeling this is going to be the biggest complain when we launch this on TestFlight... so we should do it soon

PR #2614, opened 2024-10-24
Fix profile pic banner pic upload

scoder1747:
## Summary
The existing picture-upload flow for Profile and Banner is broken. Root cause analysis of this bug is due to missing KeyPair parameter being passed while instantiating EditPictureControl objects. They KeyPair param is used to create nip98 signature and which is eventually used to create an "Authorization" header for creating an upload request.

## Checklist

- [ ] I have read (or I am familiar with) the [Contribution Guidelines](../docs/CONTRIBUTING.md)
- [ ] My PR is either small, or I have split it into smaller logical commits that are easier to review
- [ ] I have added the signoff line to all my commits. See [Signing off your work](../docs/CONTRIBUTING.md#sign-your-work---the-developers-certificate-of-origin)
- [ ] I have added appropriate `Closes:` or `Fixes:` tags in the commit messages wherever applicable, or made sure those are not needed. See [Submitting patches](https://github.com/damus-io/damus/blob/master/docs/CONTRIBUTING.md#submitting-patches)

### DIFF
--- a/damus/Views/MediaPicker.swift
+++ b/damus/Views/MediaPicker.swift
@@ -9,15 +9,19 @@ import UIKit
 import SwiftUI
 import PhotosUI
 
+enum MediaPickerEntry {
+    case editPictureControl
+    case postView
+}
+
 struct MediaPicker: UIViewControllerRepresentable {
 
     @Environment(\.presentationMode)
     @Binding private var presentationMode
+    let mediaPickerEntry: MediaPickerEntry
 
     @Binding var image_upload_confirm: Bool
-    var imagesOnly: Bool = false
     let onMediaPicked: (PreUploadedMedia) -> Void
-    
 
     final class Coordinator: NSObject, PHPickerViewControllerDelegate {
         let parent: MediaPicker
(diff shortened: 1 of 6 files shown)


jb55:
This looks right to me!

jb55:
although nostr.build supports uploads without auth right? was it not working because the image was too large?

jb55 (approved):

scoder1747:
> although nostr.build supports uploads without auth right? was it not working because the image was too large?

I thought it would support but as of now, the following is the response sent by nostr-build
Upload Error: Unauthorized, please provide a valid nip-98 token

There is one more commit that limits the profile and banner upload to image only and sets selection quantity to 1

damus/Views/MediaPicker.swift
```diff
@@ -115,9 +119,14 @@ struct MediaPicker: UIViewControllerRepresentable {
...
-        configuration.filter = imagesOnly ? .images : .any(of: [.images, .videos])
-        configuration.selection = .ordered // images are returned in the order they were selected + numbered badge displayed
+        if case .postView = mediaPickerEntry {
+            configuration.selectionLimit = 0 // Allows multiple media selection
+            configuration.filter = .any(of: [.images, .videos])
+            configuration.selection = .ordered // images are returned in the order they were selected + numbered badge displayed
+        } else if case .editPictureControl = mediaPickerEntry {
+            configuration.selectionLimit = 1 // Allows one media selection
+            configuration.filter = .images
+        }
```
jb55, line 129: please use switch statement

jb55 (changes requested):

damus/Views/MediaPicker.swift
scoder1747: @jb55 I have made the update

Issue #2766, opened 2024-12-13
cancel upload does not stop image upload


alltheseas:
reported by @scoder1747 

> when user presses the Cancel post, all the uploading operations should stop right there but currently the uploading task continues in the background with success urls
Now

jb55:
Is this still an issue?

alltheseas:
I could not recreate this issue

PR #2768, opened 2024-12-14
Fix Cancel ongoing uploading operations after user forcibly cancels the post

scoder1747:
## Summary
Issue- https://github.com/damus-io/damus/issues/2766

Bug scenario- When user presses the Cancel post, all the uploading operations should stop right there but currently the uploading task continues in the background with success urls.

Replication:
Add 5 photos from `MediaPicker` View and press 'Upload'. Then, immediately Cancel the post. Uploading task continues.

This PR introduces-
1] Cancel ongoing uploading operations after the user has pressed Cancel post 
2] Don't Generate Hepatic Feedback in this error case because user has already dismissed the `PostView`.

### Testing Performed after the changes made:
1] Add 5 photos from media pickers and press 'Upload'. Then, immediately Cancel the post. No uploading Process takes place [Passed]
2] No `UINotificationFeedbackGenerator()` is triggered [Passed]

(This change is difficult to demonstrate
on a simulator so there’s no attachment.)

## Checklist

## Test report

_Please provide a test report for the changes in this PR. You can use the template below, but feel free to modify it as needed._

**Device:** _[Please specify the device you used for testing]_

**iOS:** _[Please specify the iOS version you used for testing]_

**Damus:** _[Please specify the Damus version or commit hash you used for testing]_

**Setup:** _[Please provide a brief description of the setup you used for testing, if applicable]_

**Steps:** _[Please provide a list of steps you took to test the changes in this PR]_

**Results:**
- [x] PASS
- [ ] Partial PASS
  - Details: _[Please provide details of the partial pass]_

## Other notes

_[Please provide any other information that you think is relevant to this PR.]_

### DIFF
--- a/damus/Views/PostView.swift
+++ b/damus/Views/PostView.swift
@@ -72,6 +72,7 @@ struct PostView: View {
     @StateObject var tagModel: TagModel = TagModel()
     
     @State private var current_placeholder_index = 0
+    @State private var uploadTasks: [Task<Void, Never>] = []
 
     let action: PostAction
     let damus_state: DamusState
@@ -97,7 +98,13 @@ struct PostView: View {
 
     func cancel() {
         notify(.post(.cancel))
+        cancelUploadTasks()
         dismiss()
+    }
+    
+    func cancelUploadTasks() {
+        uploadTasks.forEach { $0.cancel() }
+        uploadTasks.removeAll()
     }
     
     func send_post() {
@@ -478,14 +485,15 @@ struct PostView: View {
                 }
                 .alert(NSLocalizedString("Are you sure you want to upload the selected media?", comment: "Alert message asking if the user wants to upload media."), isPresented: $image_upload_confirm) {
                     Button(NSLocalizedString("Upload", comment: "Button to proceed with uploading."), role: .none) {
-                        // initiate asynchronous uploading Task for multiple-images 
-                        Task {
+                        // initiate asynchronous uploading Task for multiple-images
+                        let task = Task {
                             for media in preUploadedMedia {
                                 if let mediaToUpload = generateMediaUpload(media) {
                                     await self.handle_upload(media: mediaToUpload)
                                 }
                             }
                         }
+                        uploadTasks.append(task)
                         self.attach_media = false
                     }
                     Button(NSLocalizedString("Cancel", comment: "Button to cancel the upload."), role: .cancel) {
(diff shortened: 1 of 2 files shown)


damus/Views/PostView.swift
```diff
@@ -72,6 +72,7 @@ struct PostView: View {
     @StateObject var tagModel: TagModel = TagModel()
     
     @State private var current_placeholder_index = 0
+    @State private var uploadTask: Task<Void, Never>?
```
danieldaquino, line 75: Multiple upload tasks can be started concurrently. Using a single `Task` variable causes the `cancel` to miss one of the uploads if `uploadTask` is written to more than once (e.g. selecting media from library and then pasting and image).

We should use a list, a set, or some data structure that supports multiple upload tasks

danieldaquino (changes requested):
Thank you @scoder1747 for the PR and I apologize for the delay reviewing this one.

This fix works well for cancelling a single upload task, however it has issues cancelling multiple concurrent upload operations.


**I tried the following steps:**
1. Copy an image (from anywhere) to clipboard
2. Activate "Network Link Conditioner" developer tool with a slow internet preset. (Optional, but makes the next steps easier)
2. Select 3 images from library and upload to the post view
3. While images are uploading, paste image and confirm upload.
4. Cancel post
5. Disable "Network link conditioner" (if active) and go back to Damus

**Expected result:** Haptic feedback should not occur
**Actual result:** 3 haptic feedback responses occur
**Versions used:** iPhone 13 mini, iOS 18.1, Damus `26022985a3bfb2d9dd644119b0bcebe0b9dc1976`

I believe this is because the current fix uses only one `Task` variable called `uploadTask`, so when making multiple uploads, the variable is overwritten and the previous task does not get cancelled upon post cancellation.

Maybe we should use an array or some other data structure that holds multiple tasks.

**One other minor observation:** We should add one empty line between the commit title and the rest of the commit message. I noticed git will treat the entire git commit message as the commit title unless there is a blank line between them. [Reference I found about this](https://git-scm.com/docs/git-commit/2.12.5#:~:text=Though%20not%20required%2C%20it’s%20a%20good%20idea%20to%20begin%20the%20commit%20message%20with%20a%20single%20short%20%28less%20than%2050%20character%29%20line%20summarizing%20the%20change%2C%20followed%20by%20a%20blank%20line%20and%20then%20a%20more%20thorough%20description.%20The%20text%20up%20to%20the%20first%20blank%20line%20in%20a%20commit%20message%20is%20treated%20as%20the%20commit%20title%2C%20and%20that%20title%20is%20used%20throughout%20Git.)

But otherwise this PR looks good to me! Thank you!

damus/Views/PostView.swift
scoder1747: Thanks for your feedback and the testing performed with simultaneous uploading case, @danieldaquino . I have added property `uploadTasks' for the purpose of appending the upload tasks in it in same or different uploading combinations. It will be used for canceling all the ongoing tasks. I have tested the change and it works as expected. (also added an empty line after the commit title). thanks

danieldaquino: Awesome! Thank you @scoder1747! 🙏

danieldaquino (approved):

danieldaquino:
@scoder1747, merged your changes at 282bf80daaceb775a0aca71f03cdc9dcb673be10! Thank you once again!

Issue #2819, opened 2025-01-22
Unit test failure on snapshot tests


danieldaquino:
```
  EditPictureControlTests.testEditPictureControlNotFirstTimeSetup() | damusTests/EditPictureControlTests.swift:326
      failed - No reference was found on disk. Automatically recorded snapshot: …
open "file:///Volumes/workspace/repository/damusTests/Snapshots/EditPictureControlTests/testEditPictureControlNotFirstTimeSetup.1.png"


Re-run "testEditPictureControlNotFirstTimeSetup" to assert against the newly-recorded snapshot.

  EditPictureControlTests.testEditPictureControlFirstTimeSetup() | damusTests/EditPictureControlTests.swift:298
      failed - No reference was found on disk. Automatically recorded snapshot: …

open "file:///Volumes/workspace/repository/damusTests/Snapshots/EditPictureControlTests/testEditPictureControlFirstTimeSetup.1.png"


Re-run "testEditPictureControlFirstTimeSetup" to assert against the newly-recorded snapshot.
```

For some reason the CI server is looking for snapshots in `damusTests/Snapshots`, when they are in `damusTests/__Snapshots__`, which I believe is the reason for this failure.

PR #2855, opened 2025-02-12
Dont show reposts for the same note more than once

### LINKED ISSUE #859: Consolidate reposts by multiple users for a cleaner timeline
When multiple users repost the same note, they appear individually in the timeline. If possible, a cleaner way to do this would be to combine multiple reposts that show the names of the users who all shared the same content.

Reference:
IMG_976D35D92FBF-1

jb55:
only affects home feed at the moment

### DIFF
--- a/damus/Components/Reposted.swift
+++ b/damus/Components/Reposted.swift
@@ -11,6 +11,14 @@ struct Reposted: View {
     let damus: DamusState
     let pubkey: Pubkey
     let target: NoteId
+    @State var reposts: Int
+
+    init(damus: DamusState, pubkey: Pubkey, target: NoteId) {
+        self.damus = damus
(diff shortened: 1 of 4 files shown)


jb55:
@danieldaquino I'm not sure what these failing tests are, should they be disabled ?

danieldaquino:
> @danieldaquino I'm not sure what these failing tests are, should they be disabled ?

The tests are failing because they can't find reference screenshot images. We have it tracked down here: https://github.com/damus-io/damus/issues/2819

It's likely a configuration issue on Xcode Cloud, because those images are tracked in the repo: https://github.com/damus-io/damus/tree/master/damusTests/__Snapshots__/EditPictureControlTests

PR #2928, opened 2025-03-20
Revert "Fix disappearing events on thread view"

jb55:
This commit is incorrect. We should still be using parent events from the event cache. This change introduces extreme thread loading instabilities. Revert until we have a proper fix.

This reverts commit 74d5bee1f6c4c3807fbdd41458a36dd059c896ad.


jb55:
we can re-apply this after, but I want to do a hotfix because the loading is quite bad at the moment

PR #3014, opened 2025-05-01
Translations

tyiu:


damus/Views/NoteContentView.swift
```diff
@@ -401,7 +401,7 @@ struct BlurOverlayView: View {
                     .foregroundStyle(.white)
                     .bold()
                     .padding(EdgeInsets(top: 5, leading: 10, bottom: 0, trailing: 10))
-                Text(NSLocalizedString("Media from someone you \n don't follow", comment: "Label on the image blur mask"))
```
tyiu, line 404: Arbitrary newlines in localizable strings are bad because:
1. It incorrectly makes assumptions about the size of the view and font size.
2. It's not translatable. Each language will translate this string differently, and some in ways that don't follow English grammar.

danieldaquino: Great point, thank you for catching this!

PR #3038, opened 2025-05-21
2x speed playback when you hold the screen while playing a video

YODL725:
Long-pressing videos on Damus while they play should perform a 2x speedup. Would like to see something like what has become commonplace on other platforms like X/Twitter and YouTube, as it's a huge time saver and has become expectation for many users these days.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,7 +2,7 @@
 
 # damus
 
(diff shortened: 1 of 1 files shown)


jb55:
Please open an issue instead of a PR!

PR #3137, opened 2025-07-13
Add relay hints to tags and identifiers

tyiu:
## Summary

This PR adds relay hints to tags and identifiers to improve content discovery on the network. Previously, we were referencing identifiers without them, so other clients would need to guess where to look for that content, leading to a broken user experience.

This PR does not take care of all the places that could have relay hints added. It only adds the ones where it's convenient. To fully tackle the rest, we would need a fundamental change in some of the core code.

## Checklist

- [x] I have opened or referred to an existing github issue related to this change.

## Test report

**Device:** iPhone 16 Pro Simulator

**iOS:** 18.5

**Damus:** 96cd6dc79800e7857af461085d912ea5bf2ebdd7

**Steps:**
1. Go to any note and tap on the `...` menu, and tap `Copy note ID`.
2. Observe that the clipboard has an `nevent` instead of a `note1` identifier now.
3. Paste that `nevent` identifier into search and observe that note appear.
4. Also paste it into https://njump.me/nevent... and observe that it appears there as well.
5. Tap the share button on any note and `Copy Link`. 
6. Observe that the clipboard has an `nevent`. This has not changed, but the selection of relays now prefers actual relays that we found the event from first.
7. Repeat step 5-6 but use `Share Link` instead.
8. Try replying to threads, reposting, quote reposting, reacting. You should observe that we now provide relay hints in the e and p tags.
9. For quote reposting, observe that there is a nostr:nevent... appended to the end of the note content instead of a nostr:note1...

**Results:**
- [x] PASS

### DIFF
--- a/damus/Views/PostView.swift
+++ b/damus/Views/PostView.swift
@@ -798,18 +798,18 @@ private func isAlphanumeric(_ char: Character) -> Bool {
     return char.isLetter || char.isNumber
 }
 
-func nip10_reply_tags(replying_to: NostrEvent, keypair: Keypair) -> [[String]] {
+func nip10_reply_tags(replying_to: NostrEvent, keypair: Keypair, relayURL: RelayURL?) -> [[String]] {
     guard let nip10 = replying_to.thread_reply() else {
         // we're replying to a post that isn't in a thread,
         // just add a single reply-to-root tag
-        return [["e", replying_to.id.hex(), "", "root"]]
+        return [["e", replying_to.id.hex(), relayURL?.absoluteString ?? "", "root"]]
     }
 
     // otherwise use the root tag from the parent's nip10 reply and include the note
     // that we are replying to's note id.
     let tags = [
         ["e", nip10.root.note_id.hex(), nip10.root.relay ?? "", "root"],
-        ["e", replying_to.id.hex(), "", "reply"]
+        ["e", replying_to.id.hex(), relayURL?.absoluteString ?? "", "reply"]
     ]
 
     return tags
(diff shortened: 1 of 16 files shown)


damus/Views/PostView.swift
```diff
@@ -902,15 +902,21 @@ func build_post(state: DamusState, post: NSAttributedString, action: PostAction,
...
+            content.append("\n\nnostr:\(nevent)")
 
-        tags.append(["q", ev.id.hex()]);
+            tags.append(["q", ev.id.hex(), first_relay.absoluteString, ev.pubkey.hex()]);
+            tags.append(["p", ev.pubkey.hex(), first_relay.absoluteString])
+        } else {
+            content.append("\n\nnostr:" + bech32_note_id(ev.id))
 
-        if let quoted_ev = state.events.lookup(ev.id) {
-            tags.append(["p", quoted_ev.pubkey.hex()])
```
tyiu, line 913: I actually don't understand this code. Why do a lookup again when we already have the quoted event pubkey? Seems like a bug?

damus/Views/PostView.swift
```diff
@@ -902,15 +902,21 @@ func build_post(state: DamusState, post: NSAttributedString, action: PostAction,
...
+        let relay_urls = state.nostrNetwork.relaysForEvent(event: ev)
+        if let first_relay = relay_urls.first {
+            let nevent = Bech32Object.encode(.nevent(NEvent(noteid: ev.id, relays: relay_urls.prefix(4).map { $0.absoluteString })))
+            content.append("\n\nnostr:\(nevent)")
 
-        tags.append(["q", ev.id.hex()]);
+            tags.append(["q", ev.id.hex(), first_relay.absoluteString, ev.pubkey.hex()]);
+            tags.append(["p", ev.pubkey.hex(), first_relay.absoluteString])
+        } else {
+            content.append("\n\nnostr:" + bech32_note_id(ev.id))
```
tyiu: Actually, this feels wrong for the case where we don't know which relay it came from (is this even possible? probably not, or a rare edge case when we start supporting the local relay / offline model). We can still use nevent because relays are optional but nevent also includes a pubkey. So that is strictly better than providing only a note id. I'll fix this.

damus/Views/PostView.swift
alltheseas: @danieldaquino @jb55

jb55: not sure, looks like an oversight or the result of a refactor. removing it looks fine to me

danieldaquino (approved):
Thank you @tyiu, looks good!

PR #3191, opened 2025-08-13
Update to latest nostrdb

jb55:
This should be breaking as I've changed the ndb_note_verify interface, but throwing this up here so I don't forget


danieldaquino:
@jb55, done! I rebased this and finished the ndb update on https://github.com/damus-io/damus/pull/3204

Issue #2438, opened 2024-09-07
Favorite/Starred Users


jb55:
As a Damus user who wants to have a better list experience, I would like to favourite or "star" certain users, so that I can have an easy first "Favorite follow list".

**Is your feature request related to a problem? Please describe.**

A common problem is that the people who post the most dominate the feed, while interesting people who don't post very often.

We are planning user lists:

- #531 

But this would be a very simple implementation of a list. You simply star a user to add them to your favorites. This maintains a favorite user list which is easier to get into without having to create a new list and add users to that list manually.

**Describe alternatives you've considered**

Alternatively we can build the full list UI and just use that like Xitter, but favourites would cover 90% of what most people want. I would suspect most people aren't creating many lists, they really just want a single favourite list of people they want to follow more closely.

alltheseas:
Is it a private list? 
Is it nip-51 compatible?

jb55:
@alltheseas all lists are nip-51 compatible, there is no reason to do something other than nip-51

If we make them private by default, then we won't be able to do:

- #2441 

unless we make it a private thing on the purple api.

alltheseas:
Related: receive notifications from people I subscribe to

https://github.com/damus-io/damus/issues/1265

danieldaquino:
@alltheseas @robagreda, although I imagine the UI/UX for this could be as simple as adding a star/heart button somewhere on the user profiles, I marked this one with a "design" label in case you guys have any cool ideas on how to design this special list

robagreda:
Yeah, I am on this! hehe taking time because I am designing things for notedeck but once I am ready I will tackle this! Thanks!

alltheseas:
Thanks Rob, looking forward to your masterpiece :eyes: 

> although I imagine the UI/UX for this could be as simple as adding a star/heart button somewhere on the user profiles

love the simplicity of this approach Daniel

alltheseas:
@danieldaquino using heart, instead of star, as not to confuse with purple subscription

alltheseas:
deferring post video fixes, and tap sensitivity fixes

alltheseas:
http://damus.io/figma

alltheseas:
@askeew I updated the ticket with acceptance criteria, and a link to the design system.

@jb55 @danieldaquino let me know if acceptance criteria are sufficient/too much/insufficient 🙏

danieldaquino:
> let me know if acceptance criteria are sufficient/too much/insufficient 🙏

[...]

PR #3212, opened 2025-08-29
2438 nip 81 contact card favorites

askeew:
## Summary
Implementing Favourite feed: https://github.com/damus-io/damus/issues/2438 using `NIP-81`

## Checklist

- [x] I have opened or referred to an existing github issue related to this change.

## Test report

_Please provide a test report for the changes in this PR. You can use the template below, but feel free to modify it as needed._

**Results:**
- [x] PASS
- [ ] Partial PASS


danieldaquino:
Thank you @askeew, I have not completed my review yet but so far it's looking very good! I will probably complete my review on Monday.

damus/ContentView.swift
```diff
@@ -447,6 +447,16 @@ struct ContentView: View {
         .onReceive(handle_notify(.present_full_screen_item)) { item in
             self.active_full_screen_item = item
         }
+        .onReceive(handle_notify(.contactCard)) { notifyType in
+            switch notifyType {
+            case .favorite(let pubkey):
+                damus_state.contactCards.handleFavorite(state: damus_state, target: pubkey)
```
danieldaquino: Given that:
- the `handleFavorite` method serves the purpose of formatting the new Nostr Event and sending it over the network, and
- the `notify(.contactCard(...))` calls are all coming from `ContactCardManager`

it seems that we can call `handleFavorite` methods (for both `favorite` and `unfavorite`) directly from within `ContactCardManager` (e.g. on the `toggleFavorite` method) and remove this layer of indirection.

Otherwise, this logic causes the `ContactCardManager` class to be dependent on `ContentView` for its own base functionality.

_(Note: This comment does not apply to `.favoritesUpdated` below)_

damus/Features/ContactCard/ContactCardManager.swift
```diff
@@ -0,0 +1,138 @@
...
+
+    private func createUnfavoriteContactCard(keypair: FullKeypair, target: Pubkey) -> NostrEvent? {
+        let kind = NostrKind.contact_card.rawValue
+        let tags = [
+            [Self.TARGET_PUBLIC_KEY, target.hex()]
+        ]
+        return NostrEvent(content: "", keypair: keypair.to_keypair(), kind: kind, tags: tags)
+    }
+
+    private func handlFavorite(state: DamusState, target: Pubkey, favorite: Bool) {
```
danieldaquino: (Nitpick/optional): Small typo in `handlFavorite`

damus/Features/Timeline/Views/PostingTimelineView.swift
```diff
@@ -96,6 +100,9 @@ struct PostingTimelineView: View {
                     contentTimelineView(filter: content_filter(.posts_and_replies))
                         .tag(FilterState.posts_and_replies)
                         .id(FilterState.posts_and_replies)
+                    contentTimelineView(filter: content_filter(.favorites))
```
danieldaquino: We should add a placeholder view for when the user has no favorite users setup, instead of a blank screen.

The `InnerTimelineView` might be a good place to look at to resolve this issue.

damus/Features/Timeline/Views/PostingTimelineView.swift
```diff
@@ -25,10 +25,23 @@ struct PostingTimelineView: View {
...
         var filters = ContentFilters.defaults(damus_state: damus_state)
         filters.append(fstate.filter)
+        switch timeline_source {
+        case .follows:
+            // Show content from followed users, exclude favorited users who are NOT followed
+            filters.append { ev in
+                let isFavorite = damus_state.contactCards.isFavorite(ev.pubkey)
+                // TODO: should we show also events from friendsOfFriends?
+                let isFollowed = damus_state.contacts.is_friend(ev.pubkey)
+                return !(isFavorite && !isFollowed)
```
danieldaquino: We should simply filter based on the `is_friend` result.

Currently this could match notes from unwanted users that are neither favorites nor friends.

```suggestion
                return isFollowed
```

damus/Features/Timeline/Views/PostingTimelineSwitcherView.swift
```diff
@@ -0,0 +1,42 @@
...
+
+    var body: some View {
+        Menu {
+            Picker(selection: $timelineSource) {
+                Label(TimelineSource.follows.description, image: "user-added")
+                    .tag(TimelineSource.follows)
+                Label(TimelineSource.favorites.description, image: "heart")
+                    .tag(TimelineSource.favorites)
+            } label: {
+                Text("Picker")
```
danieldaquino: It seems that this is not visible. If that is the case, perhaps we should use an empty view to avoid this word getting picked up by the translation workflow.
```suggestion
                EmptyView()
```

damus/Features/Timeline/Views/PostingTimelineSwitcherView.swift
```diff
@@ -0,0 +1,42 @@
...
+            }
+        } label: {
+            Label("", systemImage: "square.stack")
```
danieldaquino: The empty text seems not needed

We can replace this with just the image
```suggestion
            Image(systemName: "square.stack")
```

damus/Features/Timeline/Views/PostingTimelineSwitcherView.swift
```diff
@@ -0,0 +1,42 @@
...
+                .foregroundColor(DamusColors.purple)
+                .frame(width: 50, height: 35)
```
danieldaquino: The 50px width seems to be causing extra padding on the right:

Simulator Screenshot - iPhone SE (3rd generation) - 2025-09-10 at 17 52 06

Perhaps we can try doing only the height?

```suggestion
                .frame(height: 35)
```

damus/Features/Timeline/Views/PostingTimelineView.swift
askeew: This is to filter follows correctly. We don't want favorited but not followed to show up in follows timeline.

danieldaquino: If we look at the truth table for these expressions:

| Case | isFavorite | isFollowed | !isFollowed | isFavorite && !isFollowed | !(isFavorite && !isFollowed) |
|--|------------|------------|-------------|---------------------------|------------------------------|
| 1 | false      | false      | true        | false                     | **true**                     |
| 2 | false      | true       | false       | false                     | **true**                     |
| 3 | true       | false      | true        | true                      | **false**                    |
| 4 | true       | true       | false       | false                     | **true**                     |


1. The case you mention where the user is marked as "favorite" but is not "followed" _(Case 3)_ is correctly covered by the expression `return isFollowed`, because if the user is not followed, this will evaluate to `false` regardless of favorite status.
2. However, the current expression `!(isFavorite && !isFollowed)` is incorrect because it evaluates to `true` when the user is neither favorite nor followed _(Case 1)_.


In fact, the expression `!(isFavorite && !isFollowed)` is equal to `isFollowed` in all cases except _Case 1_, where the former could incorrectly display a note of someone who is not followed.

Even if _Case 1_ might be covered by other filtering elsewhere in the app and is a _"[don't-care term](https://en.wikipedia.org/wiki/Don't-care_term)"_, `return isFollowed` is simpler and faster to compute.

danieldaquino (approved):
Thank you @askeew for addressing the comments! Looks great, I can't wait to start using this feature!

Issue #3364, opened 2025-12-03
crash analysis 1.16 (1223)


alltheseas:
Crash Patterns (deduped)

  - mdb_get → ndb_lookup_by_key → Ndb.lookup_note_by_key_with_txn segfaults (6
    logs: testflight_feedback-58.zip, -68, -69, -70, -74, -75); user comments
    mention launch/boot crashes.
  - mdb_page_get → mdb_page_search_root → mdb_page_search → mdb_cursor_set
    segfaults (5 logs: -59, -60, -65, -71, -73); often when opening/profile/
    navigation.
  - RUNNINGBOARD kill (EXC_CRASH (SIGKILL), 0xdead10cc) with runloop stack (2
    logs: -57, -61); likely OS termination/background state, not a code crash.
  - mdb_txn_commit → ndb_end_query segfault (1 log: -62).
  - mdb_midl_append_list → mdb_txn_commit → ndb_end_query bus error (1 log:
    -63).
  - mdb_cursor_init → mdb_cursor_open → ndb_get_tsid segfault (1 log: -64).
  - mdb_assert_fail → mdb_page_search segfault (1 log: -67).

  All fatal crashes (except the two RUNNINGBOARD kills) are in LMDB/NostrDB
  access paths—mix of read (mdb_get/page_get/cursor) and commit (mdb_txn_commit)

alltheseas:
this PR should fix a couple of the crashes https://github.com/damus-io/damus/pull/3347

> Fixed by PR: testflight_feedback-62.zip (Pattern “mdb_txn_commit → ndb_end_query” in getDisplayName) and testflight_feedback-63.zip
    (Pattern “mdb_midl_append_list → mdb_txn_commit” in getDisplayName). The PR’s new Profiles.snapshot/ownedCopy and the switch to using
    snapshots in getDisplayName remove live NdbTxn instances from that path, so these should be resolved.

cc @danieldaquino

alltheseas:
> The big hitters are LMDB segfaults in note lookup and page search. These are in two clusters (11 of 18 logs):

>  - mdb_get → ndb_lookup_by_key → Ndb.lookup_note_by_key_with_txn (6 logs: -58,-68,-69,-70,-74,-75)
 > - mdb_page_get → mdb_page_search_root → mdb_page_search → mdb_cursor_set (5 logs: -59,-60,-65,-71,-73)

 ### What to try next (pragmatic order):

[...]

PR #3365, opened 2025-12-03
Harden LMDB note access with owned snapshots and safer logging

alltheseas:
## Summary

- 615aec84 (targets ~33% of crashes: mdb_get → ndb_lookup_by_key segfaults): add snapshot_note_by_key, prefer owned note snapshots, and tighten NdbNoteLender borrow handling to keep UI/
    async code off live NdbTxn.
  - 1a75e9f4 & e71fe438 (same ~33% mdb_get cluster): snapshot notes before streaming to async consumers; log when falling back to live transactions in subscription lookups.
  - dabbb17a (supports mdb_get/mdb_page_* segfaults ~61% overall): log failed transaction opens to expose LMDB start errors that can precede bad accesses.
  - 9857b729 / 355c8b6d (logging robustness): use StaticString formatting for the new logs above.
  - a7fc861e (guidance): document snapshot expectations for async/UI callers to avoid long-lived NdbTxn usage (addresses same mdb_get class of crashes).

### Standard PR Checklist

- [ ] I have profiled the changes to ensure there are no performance regressions, or I do not need to profile the changes.
    - Utilize Xcode profiler to measure performance impact of code changes. See https://developer.apple.com/videos/play/wwdc2025/306
    - If not needed, provide reason:
- [x] I have opened or referred to an existing github issue related to this change: https://github.com/damus-io/damus/issues/3364

## Test report
built in xcode to physical 13mini iOS 26
tinkering with app
need more testers
so far so good, no crashes

**Results:**
- [] PASS
- [x] Partial PASS
  - Details: needs testing by others

### DIFF
--- a/damus/Core/Networking/NostrNetworkManager/SubscriptionManager.swift
+++ b/damus/Core/Networking/NostrNetworkManager/SubscriptionManager.swift
@@ -317,13 +317,12 @@ extension NostrNetworkManager {
                             case .event(let noteKey):
                                 // Prefer to hand out owned snapshots so downstream async consumers never keep LMDB transactions alive.
                                 let ownedSnapshot = self.ndb.snapshot_note_by_key(noteKey, txnName: "sessionNdbStream")
-                                let lender: NdbNoteLender
-                                if let ownedSnapshot {
-                                    lender = NdbNoteLender(ownedNdbNote: ownedSnapshot)
-                                } else {
-                                    Self.logger.error("Session subscription \(id.uuidString, privacy: .public): snapshot failed for note key \(noteKey, privacy: .public)")
-                                    lender = NdbNoteLender(ndb: self.ndb, noteKey: noteKey)
+                                guard let ownedSnapshot else {
+                                    // Root cause fix: never hand out live transactions across async boundaries.
+                                    Self.logger.error("Session subscription \(id.uuidString, privacy: .public): snapshot failed for note key \(noteKey, privacy: .public); dropping to avoid leaking live txn")
+                                    continue
                                 }
+                                let lender = NdbNoteLender(ownedNdbNote: ownedSnapshot)
                                 try Task.checkCancellation()
                                 guard let desiredRelays else {
                                     continuation.yield(.event(lender: lender))  // If no desired relays are specified, return all notes we see.
@@ -370,11 +369,11 @@ extension NostrNetworkManager {
             // Since note ids point to immutable objects, we can do a simple ndb lookup first
             if let noteKey = self.ndb.lookup_note_key(noteId) {
                 // Snapshot eagerly to avoid passing live transactions to async callers.
-                if let owned = self.ndb.snapshot_note_by_key(noteKey, txnName: "SubscriptionManager.lookup") {
-                    return NdbNoteLender(ownedNdbNote: owned)
+                guard let owned = self.ndb.snapshot_note_by_key(noteKey, txnName: "SubscriptionManager.lookup") else {
+                    Self.logger.error("Lookup snapshot failed for note key \(noteKey, privacy: .public); dropping to avoid leaking live txn")
+                    return nil
                 }
-                Self.logger.error("Lookup fallback to live txn for note key \(noteKey, privacy: .public)")
-                return NdbNoteLender(ndb: self.ndb, noteKey: noteKey)
+                return NdbNoteLender(ownedNdbNote: owned)
             }
             
             // Not available in local ndb, stream from network
--- a/nostrdb/UnownedNdbNote.swift
+++ b/nostrdb/UnownedNdbNote.swift
@@ -8,6 +8,8 @@
 /// Allows an unowned note to be safely lent out temporarily.
 ///
 /// Use this to provide access to NostrDB unowned notes in a way that has much better compile-time safety guarantees.
+/// Important: do not pass lenders carrying live transaction lookups across async boundaries—callers should snapshot
+/// or copy at the boundary so no live LMDB transaction escapes its scope.
 ///
 /// # Usage examples
 ///
@@ -122,6 +124,8 @@ enum NdbNoteLender: Sendable {
     }
 
     /// Materialize an owned snapshot of the note, avoiding any long-lived LMDB transaction usage.
+    /// Unlike `getCopy()`, this bypasses the regular borrow/toOwned path and uses `snapshot_note_by_key`
+    /// so callers never touch a live LMDB transaction before getting the owned copy.
     func snapshot() throws -> NdbNote {
         switch self {
         case .ndbNoteKey(let ndb, let noteKey):


alltheseas:
Screenshot 2025-12-03 at 11 56 26 AM

nostrdb/UnownedNdbNote.swift
```diff
@@ -106,14 +112,27 @@ enum NdbNoteLender: Sendable {
...
+        switch self {
+        case .ndbNoteKey(let ndb, let noteKey):
+            if let owned = ndb.snapshot_note_by_key(noteKey) {
+                return owned
+            }
+            throw LendingError.errorLoadingNote
+        case .owned(let note):
+            return note.to_owned()
+        }
+    }
```
danieldaquino, line 139: Isn't this equivalent to the `getCopy` method? What is the difference, functionally speaking? And what is the reason that we need both?

damus/Core/Networking/NostrNetworkManager/SubscriptionManager.swift
```diff
@@ -315,7 +315,15 @@ extension NostrNetworkManager {
...
-                                let lender = NdbNoteLender(ndb: self.ndb, noteKey: noteKey)
+                                // Prefer to hand out owned snapshots so downstream async consumers never keep LMDB transactions alive.
+                                let ownedSnapshot = self.ndb.snapshot_note_by_key(noteKey, txnName: "sessionNdbStream")
+                                let lender: NdbNoteLender
+                                if let ownedSnapshot {
+                                    lender = NdbNoteLender(ownedNdbNote: ownedSnapshot)
+                                } else {
+                                    Self.logger.error("Session subscription \(id.uuidString, privacy: .public): snapshot failed for note key \(noteKey, privacy: .public)")
+                                    lender = NdbNoteLender(ndb: self.ndb, noteKey: noteKey)
+                                }
```
danieldaquino, line 324: This approach feels a bit like avoiding the problem rather than solving it, and it seems to negate one of the benefits of using NostrDB.

However, I am willing to go for it as a temporary fix if it effectively reduces the crash incidence and performance feels acceptable. But in that case we should add a `TODO` comment indicating that eventually we want to avoid copying notes and allow callers to borrow to allow for performance optimizations.

damus/Core/Networking/NostrNetworkManager/SubscriptionManager.swift
```diff
@@ -361,6 +369,11 @@ extension NostrNetworkManager {
             
             // Since note ids point to immutable objects, we can do a simple ndb lookup first
             if let noteKey = self.ndb.lookup_note_key(noteId) {
+                // Snapshot eagerly to avoid passing live transactions to async callers.
+                if let owned = self.ndb.snapshot_note_by_key(noteKey, txnName: "SubscriptionManager.lookup") {
+                    return NdbNoteLender(ownedNdbNote: owned)
+                }
+                Self.logger.error("Lookup fallback to live txn for note key \(noteKey, privacy: .public)")
```
danieldaquino: This approach feels a bit like avoiding the problem rather than solving it, and it seems to negate one of the benefits of using NostrDB.

However, I am willing to go for it as a temporary fix if it effectively reduces the crash incidence and performance feels acceptable. But in that case we should add a TODO comment indicating that eventually we want to avoid copying notes and allow callers to borrow to allow for performance optimizations.

nostrdb/UnownedNdbNote.swift
alltheseas: > On getCopy() vs snapshot(): Both give you an owned note, but they take different routes. getCopy() briefly opens a live database transaction to borrow the note and then copies it. snapshot() skips that and asks the database for a ready-made owned copy, so callers never touch a live transaction. We keep both because some call sites only have a key (where snapshotting is safer for async use), while others already hold a note and just need a copy.

damus/Core/Networking/NostrNetworkManager/SubscriptionManager.swift
alltheseas: this is great feedback. proposed path:

> To keep borrowing and performance while fixing the crashes:

>  - Keep borrows short and synchronous: do the work while you’re still in the same call stack as the borrow; don’t pass the borrower into async
    tasks or store it.
>  - Cut async boundaries: if you must go async, turn the note into a safe form right there (e.g., minimal fields you need), and don’t carry the
    live transaction across.
>  - Push processing up front: do the heavy work at the point of borrow, hand downstream only the results, not the borrower.
>  - Add guardrails: assertions/logging when a borrower is used across an async hop or stored for later; tests that ensure borrowers aren’t
    retained beyond the scope.
>  - Profile then optimize: once lifetimes are enforced, we can keep borrowing in hot paths that stay synchronous and only copy where absolutely
    necessary.

>  This keeps the performance benefit of borrowing in safe, short-lived paths and eliminates the crash vector by never letting live transactions
  escape.

damus/Core/Networking/NostrNetworkManager/SubscriptionManager.swift
alltheseas: great feedback, left comment below

damus/Core/Networking/NostrNetworkManager/SubscriptionManager.swift
alltheseas: I will have a go

alltheseas:
@danieldaquino in https://github.com/damus-io/damus/pull/3365/commits/0d129503ae57757c201159246212baefc9f0eea7 had a go at responding to your feedback:

> Lookup path: We no longer fall back to a live transaction. If snapshot_note_by_key fails, we log and return nil instead of handing out a
    live borrower. That removes the crash vector from async use of live LMDB transactions, rather than “avoiding the problem.”
>  - Subscription path: Likewise, we no longer wrap a live transaction when snapshotting fails. We log and drop the event so nothing downstream
    touches a live txn across async boundaries. Again, this eliminates the root cause (leaking live txns), instead of papering it over.

  >Tradeoff: This keeps safety but can drop events if snapshotting fails, and we lose the performance benefit of borrowing in these async paths.
  If we need borrowing/perf later, we’ll need a design where borrows stay strictly synchronous and never cross async boundaries; for now,
  snapshot-only stops the crashes without TODO debt.

alltheseas:
I have not been able to successfully conduct performance profiling on my machine. I did give it a go a few times 😕. After not / crashing, this is important to test.

danieldaquino:
> Lookup path: We no longer fall back to a live transaction. If snapshot_note_by_key fails, we log and return nil instead of handing out a
live borrower. That removes the crash vector from async use of live LMDB transactions, rather than “avoiding the problem.”

1. `NdbNoteLender` only stores either a note key (which is simply a number) for later lookup and borrowing, or an owned copy of `NdbNote`. It doesn't "hand out" live LMDB transactions. These assertions are completely inaccurate.
2. It does "avoid the problem" to the extent that it performs a full `NdbNote` memory copy operation regardless of whether the caller needs it or not. As I mentioned, this might be an ok tradeoff for now, but eventually we want to allow the caller to borrow from Ndb. In the long run the way we prevent dangerous async use of LMDB transactions is by making use of new Swift `~Copyable` + actor + borrow semantics that prevent such use at the compiler level.

damus/Core/Networking/NostrNetworkManager/SubscriptionManager.swift
danieldaquino: Where is the comment?

danieldaquino:
Despite some issues I have with the AI's analysis, there are some good ideas here that I think it's worth exploring, even if only as a temporary fix. I will push an internal TestFlight with this for more testing.

danieldaquino:
Placed this PR on the queue for now to prioritize other PRs while we decide whether to merge https://github.com/damus-io/damus/pull/3366, which (if merged) might require some changes to this PR. I will come back to this soon.

alltheseas:
Closing in favor of consolidated Linux-style topic branches. This work will be audited against AGENTS.md requirements and cherry-picked into ordered topic branches. See https://github.com/damus-io/damus/issues/3624 for tracking.

PR #3626, opened 2026-02-14
SwiftUI performance optimizations

jb55:
## Summary

Comprehensive SwiftUI performance overhaul targeting main-thread hitches during scrolling, profile viewing, and video playback. 20 commits across 19 files, each addressing a specific bottleneck identified via Instruments profiling.

### View tree & layout optimizations
- **Remove AnyView type erasure from hot render paths** — EventActionBar, EventView, EventShell, EventMutingContainerView, and ProfileView banner all replaced AnyView with `@ViewBuilder` or generics, restoring SwiftUI's structural identity diffing instead of full subtree teardown/rebuild
- **Flatten Route.view()** — The 44-case `@ViewBuilder` switch created 44 levels of nested `_ConditionalContent`; replaced with explicit `AnyView` returns (appropriate here since navigation destinations are one-shot, not per-frame)
- **Remove unnecessary wrapper containers** — VStack, Group wrappers in the event render path that added layout nodes with no purpose
- **Reduce GeometryReaders** — ChatroomThreadView reduced from 5 to 3 GeometryReaders; TimelineView eliminated a redundant one by folding into an existing scroll callback
- ~~Replace paging TabView~~ — PostingTimelineView's `.page` style TabView was pre-rendering both tabs (~129ms for the offscreen one); replaced with a simple conditional (REMOVED)

### Caching & allocation reduction
- **Cache DateComponentsFormatter** — Was allocating a new formatter per-note per-render (8ms in profiler); now a static singleton
- **Cache locale bundles** — `bundleForLocale` was hitting disk (Bundle.main.path + Bundle(path:)) 12+ times per ProfileView render; now cached by locale identifier
- **Cache contact pubkeys** — `follows()` was O(n) linear scan of contact event tags; now O(1) Set lookup built once on contact event arrival
- **Pre-render FriendIcon gradient** — The gradient-masked SF Symbol was forcing offscreen compositing every frame (~8.8ms/frame); now blits a pre-rendered UIImage
- **Defer FollowingModel construction** — Was iterating all contact tags on every ProfileView.body; now builds lazily at navigation time

### ObservableObject / state invalidation fixes
- **Stop ProfileModel.objectWillChange on every event** — EventHolder already notifies InnerTimelineView independently; the redundant notification was causing full ProfileView re-evaluation (~152ms) on every network event
- **Stop DamusVideoPlayer.current_time from triggering objectWillChange** — The per-second timer was re-evaluating every visible video player's full body; replaced `@Published` with `CurrentValueSubject` scoped to controls only
- **Isolate scroll-driven updates in ProfileView** — Replaced `@State var yOffset` with a reference-type `ScrollOffsetTracker` so scroll mutations don't trigger parent body re-evaluation (~112ms saved per frame)
- **Use reference-type storage in OffsetHelper** — `@State` value types were causing body re-evaluation on every scroll frame
- **Extract ProfileAboutSection** — Was a computed property rebuilt on every body evaluation, doing NDB lookups and name resolution each time

### UIKit drop-ins
- **Replace SwiftUI Menu with UIKit UIButton** — SwiftUI Menu eagerly evaluates its content builder on every row (15ms+); UIButton with `UIDeferredMenuElement.uncached` defers to tap time
- **Always downsample images to target size** — Images under the size threshold were going through deferred JPEG decode, causing 12ms+ main-thread decode in `CA::Layer::prepare_commit`; now all images decode at target resolution via `CGImageSourceCreateThumbnailAtIndex`
- **Skip redundant AVPlayerViewController.player assignment** — `setPlayer` rebuilds internal video layers (~13ms) even when assigned the same reference; added identity guard

## Test plan
- [ ] Scroll main timeline — verify smooth scrolling, no visual regressions in note cells (action bars, menus, friend icons, timestamps)
- [ ] Open profile view — verify banner, pfp, about section, follow counts render correctly; scroll up/down and verify parallax/toolbar transitions
- [ ] Tap "Following" on a profile — verify follower list loads correctly
- [ ] Play a video in timeline — verify controls (play/pause, timestamp) work; verify no unnecessary flickering
- [ ] Open chatroom thread — verify thread lines and scroll behavior
- [ ] Test "Notes" / "Notes & Replies" tab switching on posting timeline
- [ ] Verify context menus on notes still work (now UIKit-based)
- [ ] Profile images and banners load and display at correct resolution

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Release Notes

* **Performance Improvements**
  * Optimized video playback time synchronization for smoother scrubbing and display updates.
  * Enhanced profile page rendering to reduce lag during scrolling.
  * Pre-rendered profile icons for faster display.
  * Improved image caching and SVG detection to speed up media loading.

* **Bug Fixes**
  * Fixed image downsampling to consistently apply across all data types.

### DIFF
--- a/damus/Features/Profile/Views/ProfileView.swift
+++ b/damus/Features/Profile/Views/ProfileView.swift
@@ -216,188 +216,17 @@ struct ProfileView: View {
         }
     }
 
-    func lnButton(profile: Profile?, lnurl: String?) -> some View {
-        return ProfileZapLinkView(profile: profile, lnurl: lnurl, profileModel: self.profile) { reactions_enabled, lud16, lnurl in
-            Image(reactions_enabled ? "zap.fill" : "zap")
-                .foregroundColor(reactions_enabled ? .orange : Color.primary)
-                .profile_button_style(scheme: colorScheme)
-                .cornerRadius(24)
-        }
-    }
-    
-    var dmButton: some View {
-        let dm_model = damus_state.dms.lookup_or_create(profile.pubkey)
-        return NavigationLink(value: Route.DMChat(dms: dm_model)) {
-            Image("messages")
-                .profile_button_style(scheme: colorScheme)
-        }
-    }
-    
-    private var followsYouBadge: some View {
-        Text("Follows you", comment: "Text to indicate that a user is following your profile.")
-            .padding([.leading, .trailing], 6.0)
-            .padding([.top, .bottom], 2.0)
-            .foregroundColor(.gray)
-            .background {
-                RoundedRectangle(cornerRadius: 5.0)
-                    .foregroundColor(DamusColors.adaptableGrey)
-            }
-            .font(.footnote)
-    }
-
-    func actionSection(ndbprofile: Profile?, lnurl: String?, pubkey: Pubkey) -> some View {
-        return Group {
-            if damus_state.settings.enable_favourites_feature {
-                FavoriteButtonView(pubkey: profile.pubkey, damus_state: damus_state)
-            }
-            if let profile = ndbprofile,
-               let lnurl,
-               lnurl != ""
-            {
-                lnButton(profile: ndbprofile, lnurl: lnurl)
-            }
-
-            dmButton
-
-            if profile.pubkey != damus_state.pubkey {
-                FollowButtonView(
-                    target: profile.get_follow_target(),
-                    follows_you: profile.follows(pubkey: damus_state.pubkey),
-                    follow_state: damus_state.contacts.follow_state(profile.pubkey)
-                )
-            } else if damus_state.keypair.privkey != nil {
-                NavigationLink(value: Route.EditMetadata) {
-                    ProfileEditButton(damus_state: damus_state)
-                }
-            }
-
-        }
-    }
-
-    func nameSection(ndbprofile: Profile?, lnurl: String?) -> some View {
-        return Group {
-            let follows_you = profile.pubkey != damus_state.pubkey && profile.follows(pubkey: damus_state.pubkey)
-
-            HStack(alignment: .center) {
-                ProfilePicScrollEffect(
-                    scrollTracker: scrollTracker,
-                    pubkey: profile.pubkey,
-                    pfp_size: pfp_size,
-                    navbarHeight: navbarHeight,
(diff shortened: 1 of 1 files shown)


danieldaquino:
@jb55, can you please add sign-offs and changelog entries on the commits?

jb55:
the commits are fine, we don't need to be so anal about this

danieldaquino:
> the commits are fine, we don't need to be so anal about this

Should we make the sign-off requirement optional in the contribution guidelines?

danieldaquino:
There is an issue with the profile view banner heading:
IMG_0415


Expected behaviour (Taken from 1.17 before the PR):

Screenshot 2026-02-18 at 19 42 03


Other elements seem to be working as expected.

danieldaquino (changes requested):
Thanks @jb55, changes make sense and views seem to be working after changes. There only seems to be one new issue (as far as I have tested) with the profile banner top bar that we should address before merging (More details on the previous comment).

danieldaquino:
Submitted to TestFlight for testing with users. Version and build number: v1.17 (1299)

danieldaquino:
One issue raised by a user: The home view does not scroll/swipe between Notes and Replies and Notes. 

Quote: 
> Oh wow, please fix this. I’m so used to sliding between notes / notes and replies.

@jb55, is it feasible to maintain that behaviour alongside the relevant performance improvement?

jb55:
Do people use that? Apparently its a huge performance issue because it actually renders both views

jb55:
I dropped that commit and rebased

jb55:
> > the commits are fine, we don't need to be so anal about this
> 
> Should we make the sign-off requirement optional in the contribution guidelines?

Yeah

jb55:
apparently i pushed github/experimental-pr-3626 instead of jb55/opus-perf, not sure why there are two branches for this

jb55:
the visual glitch on profile is definitely annoying, but i want to fix the architectural perf issues first and fix those later

danieldaquino:
> the visual glitch on profile is definitely annoying, but i want to fix the architectural perf issues first and fix those later

Have you had a chance to fix these? Please let me know how to proceed with this PR.

